PR #77, opened 2012-07-05
Config option: Fillers destroy blocks

Maeyanie:
Add config option to support Fillers destroying blocks instead of breaking them.
Prevents Fillers from being (ab)used as cheap Quarries.

By popular request. :)


AlexIIL:
@Maeyanie Please could you take a look at the [relicensing issue](https://github.com/BuildCraft/BuildCraft/issues/3004) for this?

Maeyanie:
I don't see my name in the list there, but sure, I'll post just in case.

AlexIIL:
Thanks :)

(A lot of commits aren't associated with a github user, so the original lists didn't find them)

PR #2377, opened 2015-01-11
Make swimming in oil a lot harder

Vexatos:
Swimming upwards is pretty much impossible in Vanilla with this (some mods might make it possible).
Crude Oil is a very dense semifluid, you should not be able to swim properly in it; if you jump into an oil well, you will most likely die but that's your own fault then. Oil is deadly.
Also added a config option to restore previous behaviour.


bochen415:
Thanks guys.
Now I feel great swimming in oil
6a00d83451b96069e201156e4ffd82970c-400wi

PR #2390, opened 2015-01-17
Fast auto workbench

immibis:
This adds a config option "autoworkbench.fast", which makes auto-workbenches able to craft every tick (instead of every 256).

Also included is an optimization to make auto-workbenches not retry crafting unless the input grid has changed, and to make them not rebalance every tick (they rebalance when the input grid has changed since the last rebalance, or if the last rebalance hasn't finished yet).


asiekierka:
Merged partially. The optimizations are nice, however there are enough auto workbenches in other mods for me not to see a direct need for a config option for ours.

PR #2431, opened 2015-02-01
Bugfixes

warlordjones:


AEnterprise:
@warlordjones you broke formatting, changing lines and the code style that din't need to changed, could you please fix the style so we can easely see what you've done?

warlordjones:
Bah, that would be my automatic codestyler, sorry!

warlordjones:
That should be only the changes, minus autostyling now. Derp. I'm not readding the robot thing, as I misinterpreted what the issue was talking about.

warlordjones:
Well, mostly minus autostyling, I can't seem to stop it removing empty lines and replacing them with...empty lines?

AEnterprise:
having the refinery only accept oil is a bad idea, other mods can add refinery recipes with the api, you could make it only accept fluids that are part of a valid recipe

warlordjones:
Ok, I'll do that

warlordjones:
Hmm, this requires quite some changes, to allow multiple accepted fluids in the SingleUseTank.

AEnterprise:
not realy, you can also perform the check here: https://github.com/BuildCraft/BuildCraft/blob/6.3.x/common/buildcraft/factory/TileRefinery.java#L298

warlordjones:
Ah, that could be better

warlordjones:
That should work

warlordjones:
Most recent changes should help reduce BCs currently massive network usage. Not fully tested, so it may break some stuff.

asiekierka:
Augh. I'll have to read this all in a few days.

warlordjones:
HMm?

asiekierka:
It's a lot of commits. I generally review commits before merging.

warlordjones:
Ah, right, yeah. Makes sense

warlordjones:
The first four are derps!

psxlover:
You could (and should) squash them.

warlordjones:
Ah, yes, I should do that

common/buildcraft/core/TileBuildCraft.java
```diff
@@ -49,10 +46,12 @@
 	private String owner = "[BuildCraft]";
 	private RFBattery battery;
 
+	private SafeTimeTracker updateNetworkTime = new SafeTimeTracker(BuildCraftCore.updateFactor - 10);
```
asiekierka: Why is this parameter here, and why is it "- 10"?

asiekierka:
I refuse to merge in the following changes:
1. Changing the default update factor. It's slow enough already. If you want something that updates twice as slow, multiply the factor by two on THAT side.
2. "sendNetworkUpdateEveryTick()". Again, tuning the network updates should be done by the machine tiles - BC's main tile should enforce NO THROTTLING.

warlordjones:
Ok, I'll change that

Arona
[+] You're welcome to add me to your address book
<https://ws.writethat.name/kwaga-bin/titan/WEB/me.pl/2040834221374130777/i>

warlordjones:
History rewritten. Commits squashed.

warlordjones:
That...would...be better.

warlordjones:
The issue is that there are multiple ways to set, and therefore to get, block drops. I think this is the one that covers the most bases.

asiekierka:
@warlordjones The change is incorrect with the BlockUtils thing. Your code does not take into account many things. What I think we should do is ask @RWTema what DenseOres does different. So, please revert that.

Sorry. Other than that, it's 100% OK. Really this time.

psxlover:
```java
+ if(RefineryRecipeManager.INSTANCE.getValidFluidStacks1().contains(resource))
used += tanks[0].fill(resourceUsing, doFill);
resourceUsing.amount -= used;
+ if(RefineryRecipeManager.INSTANCE.getValidFluidStacks2().contains(resource))
used += tanks[1].fill(resourceUsing, doFill);
```
Fix the indentation of this code.

Also of-topic but shouldn't the updateRecipe() only get called if the liquid in one of the tanks changes? No need to reset it every tick.

asiekierka:
@psxlover is right - the update should only be called upon call of a fill or drain, or after the refinery carries out a processing step.

asiekierka:
But we'll leave that for another PR. Enough on this guy.

warlordjones:
Phew! Good to see you keep strict standards :P

PR #2435, opened 2015-02-02
fix an error when loading old robots due to package renames

hea3ven:
Might be worth to note in the release that if you upgrade to 6.4.x and then downgrade, the robots won't load and you will lose them.


asiekierka:
Could you repeat what I did in 6.3.x for pipe pluggables, that is make each robot AI have its own name which is stored in a registry and NOT use class names?

hea3ven:
yeah, sure.

asiekierka:
Closed until that is done.

PR #2447, opened 2015-02-07
fix the architect table sometimes modifiying the world when creating blu...

### LINKED ISSUE #2424: Architect Table crash + Builder unable to place chests with signs on
Mods: BC 6.3.3, BCA 1.8.4, Openblocks, NEI, WAILA

Firstly I crashed When trying to place a blank blueprint in an architect table : http://pastebin.com/rcQgj4G5

This crash happened when I changed the option in the architect table from survival to creative and put a blank blueprint into it.

I already have a blueprint of the build and when i tested buiding it using the builder, It doesn't place the chests and the signs that are supposed to be on the chests just fall on the ground and the chests aren't placed

Blueprint + World download: https://drive.google.com/folderview?id=0B0kT-t-Y3RDTYlg1Z3hGSFlVX0U&usp=sharing 

Images: http://imgur.com/a/UFRvB

hea3ven:
...eprints, fixes #2424

### DIFF
--- a/api/buildcraft/api/blueprints/SchematicTile.java
+++ b/api/buildcraft/api/blueprints/SchematicTile.java
@@ -72,6 +72,7 @@ public class SchematicTile extends SchematicBlock {
 			if (tile != null) {
 				tile.writeToNBT(tileNBT);
 			}
+			tileNBT = (NBTTagCompound) tileNBT.copy();
 		}
 	}
 


asiekierka:
This fixes many more things than just that, and would be better suited for the 6.4.x branch as it touches on the API, however I don't plan more 6.3.x releases so whatever.

PR #2448, opened 2015-02-07
Wix API-core dependencies

viliml:
The API package mustn't depend on the core, in order or it to be independent.
To do that, I extracted some methods from Utils, and moved the Block Property declarations to a new class in the API, so classes in the API, and addon mods too, can use them.


asiekierka:
Why 1.8? The 1.8 branch is currently dead because it's outdated. Go for 6.4.x.

asiekierka:
Oh, this is 1.8 specific. Alright.

PR #2459, opened 2015-02-13
Stripes refactor

hea3ven:


api/buildcraft/api/transport/IStripesHandler.java
```diff
@@ -17,6 +17,7 @@
 public interface IStripesHandler {
 	public static enum StripesHandlerType {
 		ITEM_USE,
+		ITEM_USE_LAST,
```
asiekierka: I don't think having a publicly available ITEM_USE_LAST is a good idea. Consider making StripesHandlerDefault added to the end of the list of all stripes handlers from the handling code.

asiekierka:
The rest looks good to me!

api/buildcraft/api/transport/IStripesHandler.java
hea3ven: Yeah, I agree. The other option I considered is splitting handlers into two lists, high priority handlers and low priority handlers.

asiekierka: Ping?

PR #2468, opened 2015-02-17
fix direction parameters choosing the wrong directions, fix #2454

### LINKED ISSUE #2454: Gate actions with parameter
I can choose parameter only Pipe disconnecting side on Gate Actions (Closed, Open, Input Only, Output Only) in Quartz and Emerald Gate

hea3ven:


asiekierka:
Great!

Also, note for your stripes pipe rework - could you look into a certain highly requested feature? (void pipe in stripes pipe retracts the pipe by 1 block instead of extending it, making both the void pipe and removed pipe (and all its drops) come back)

PR #2473, opened 2015-02-18
Update LICENSE.API

### LINKED ISSUE #2470: Incorrect Copyright Year
Needs to be changed from 2014 to 2015.

Techokami:
Fixes #2470

### DIFF
--- a/LICENSE.API
+++ b/LICENSE.API
@@ -1,6 +1,6 @@
 BuildCraft API License
 
-Copyright (c) 2011-2014 SpaceToad and the BuildCraft team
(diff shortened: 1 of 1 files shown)


hea3ven:
It's missing the guidelines/header.txt file, and all the source files' headers. It would probably be wise to make a bash script, or a gradle task that does this automatically, so that we don't have to worry about this again next year.

Techokami:
welp;

Kubuxu:
There is gradle plugin for that. It normally checks for headers but have also option to add them all.

Issue #2045, opened 2014-10-18
Make BuildCraft splittable again


asiekierka:
Module | Responsibilities | Dependencies | Done
------ | ------ | ------ | -------
Core | Provide core functionality shared between multiple modules, few actual blocks/items (wrench, map location, ...) | - | -
Silicon | Laser tables | Core | Yes
Transport | Pipes of all kinds, Gates, Facades | Core, Silicon for gate recipes | Yes
Robotics | Robots, Redstone Boards, Zone Planners | Core, Silicon, Transport | Yes
Factory | Machines, quarries, refineries, mining wells | Core | Yes
Builders | Builder, filler, blueprints, schematics | Core | Yes
Energy | Engines, fuel, oil deposits | Core | Yes

asiekierka:
And with the 6.5.x branch, I'm almost done!

liach:
How about API?

PR #2474, opened 2015-02-18
Can to disable mods

progre:
There is a lot of blocks and items on BuildCraft.
I think users can select needful mod.

Example:
(main.conf)
    S:disabledMods <
        "BuildCraft|Energy"
        "BuildCraft|Silicon"
        "BuildCraft|Transport"
     >

--> User can use core, builders, and factory.
    This is useful when using with EnderIO and others.

### DIFF
--- a/api/buildcraft/api/boards/RedstoneBoardRegistry.java
+++ b/api/buildcraft/api/boards/RedstoneBoardRegistry.java
@@ -9,13 +9,14 @@
 package buildcraft.api.boards;
 
 import java.util.Collection;
+import java.util.Collections;
 
 import net.minecraft.client.renderer.texture.IIconRegister;
 import net.minecraft.nbt.NBTTagCompound;
 
 public abstract class RedstoneBoardRegistry {
 
-	public static RedstoneBoardRegistry instance;
+	public static RedstoneBoardRegistry instance = new NullRedstoneBoardRegistry();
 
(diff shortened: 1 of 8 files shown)


AEnterprise:
a better thing would be to make BC splitable again like it used to be

progre:
It's so good. However, I think it isn't be easy because dependency is strong.
I think so, and did a minimal change.

AEnterprise:
currently there is indeed a strong dependency, @asiekierka said earlier he was working on making BC splittable again and i think that's a better solution

just my personal opinion though

progre:
Oh, it's so pretty! (I found it in #2045)
I think I should wait that issue so I close this pullreq. Thx for opinion!

AEnterprise:
@progre asie is no longer the main developer, he quit modding so if you want to do make BC splitable i think you are free to work on that

progre:
Hm, It's interesting.
I tentatively see that issue.

asiekierka:
Splitting BuildCraft is not as easy as you'd like to think. There are many cross-dependencies that are fragile to touch.

asiekierka:
I might finish it when I get some free time. What would be easier is adding a simple way to enable and disable items and blocks - see CoreProxy.register* and add hooks there.

progre:
ok. I leave not touch if you say so.

PR #2479, opened 2015-02-20
I'm an idiot *facepalm*

viliml:
So, apparently, my *fix* for API dependencies introduced MORE invalid dependencies... That was a fun time!

Also removed some weird sun internal imports that weren't used for anything and just caused problems.


viliml:
Anyway, now the enums used by BC are available in the API for addons to detect and use and stuff... Hopefully it proves useful or something...

Thog:
Can you squash your commits?

viliml:
Well, this didn't work...

PR #2480, opened 2015-02-21
Finally really fix API dependencies, after many derps

viliml:
So, apparently, my *fix* for API dependencies introduced **MORE** invalid
dependencies... That was a fun time!

Also removed some weird com.sun.xml.internal.ws.util imports that
weren't used for anything and just caused problems.

I also fixed formatting in the api package, and that's when I realized
there's hundreds more style errors in the main code...

**And to top it all off, _I can't use Git at all!_**


AEnterprise:
git tutorials: http://gitready.com/

viliml:
Yeah, I tried that. I used rebase to squash the commits, but when I synced it they came back. Also, push didn't work and asked me to pull, and pull made more errors...

AEnterprise:
if you squash commits that you already pushed to github you need to force the push (-f ) to override the commits that you already pushed

viliml:
I really should have thought of that *sigh*... Anyway, that worked, thanks!
Anything else I need to change here?

PR #2483, opened 2015-02-22
Update to latest BC 6.2

viliml:
May be some rendering bugs, not more than before.


Thog:
The mc1.8 branch is already in 6.2, Why does it need to be updated? The first thing that is needed is to finish the 1.8 port AND after that merge the 6.2.X on it. Also, you enter in conflict with my latest commits. For me, this PR isn't actually needed.

Best Regard, Thog.

asiekierka:
No. Not yet. This is too early. Leave the 1.8 port to Thog - it's VERY low priority.

PR #2543, opened 2015-03-08
fix planters not working with zone planner areas

hea3ven:


asiekierka:
Is this really happening?

Issue #2372, opened 2015-01-10
Clay Pipe with Filter 


tokoku-11:
Another color items don't pass Clay Pipe with Filter 

![2015-01-10_12 01 05](https://cloud.githubusercontent.com/assets/7271486/5690253/d9386aa8-98c0-11e4-846b-e23683385802.png)

asiekierka:
...oh. This is a very tough bug to solve, as it essentially requires the rewrite of the entire lens/filter system. For this, you might find it is better to use regular pipes with filters on them and not clay pipes.

Kubuxu:
This is real issue as this feature is really useful.

asiekierka:
@Kubuxu - regular stone pipes will work in the same way with filters.

Kubuxu:
Will they prioritize inventories?
And will they prioritize inventories if there are two filters of same color but one toward inventory.

asiekierka:
@Kubuxu - No, I'm afraid, but those are much rarer usecases. If you look into the source code, you will see why it is so hard to do with the current system.

viliml:
Wouldn't it work if the pipes' event handlers were a lower priority than the pluggables'?
That way the clay pipe won't try to push the wrongly colored items through the filter.

asiekierka:
@viliml - that causes other pipes to break. I think a priority system would help?

viliml:
Well, either a priority system, or "pre" and "post" sub-events for FindDest.
Are you sure making pluggables process before pipes would cause some of them to break? The only handlers using that event besides Filters are Clay, Diamond and Diazuli pipes, so the only change would be ensuring the filter is run first.
1. In the case of the Clay pipe, it would remove ones that didn't match the color, and specifically fix this problem
1. In the case of the Diazuli pipe

viliml:
Oh, right, I see what you meant. There are other events and some of them would break if pluggables were ran first? But even then, the behavior right now is random, so adding *any* kind of order shouldn't break it...
Either way, The PipeEventItem.FindDest event could be split into PlugDest(pluggables would remove certain sides from the list of possible destination) and ChooseRoute(the pipe would choose which of the remaining possible destinations to send the item). Both names subject to change.

asiekierka:
@viliml - actually, the order is not random right now. The event bus is registered on in order.

What I plan to do is add @Priority(Priority.LOW/NORMAL/HIGH)

viliml:
In order of registration huh...

[...]

PR #2545, opened 2015-03-09
Filter fix

viliml:
For https://github.com/BuildCraft/BuildCraft/issues/2372 .
Tested and confirmed to solve the problem. If the event bus truly handles events in order of registration, it should work fine.


asiekierka:
It was already this way during early 6.3.0 testing, but it broke other things - can't remember what, so I swapped it as the swap made only the clay pipe break.

PR #2552, opened 2015-03-12
Make slimeball recipe work with oredict. Close #2551.

### LINKED ISSUE #2551: oredict slime balls recipe
The slime balls -> pipe sealant should use ore dict to allow slime balls from other mods to be used.

Kubuxu:


Kubuxu:
Or should I add it to 6.5.x branch?

asiekierka:
What, from other mods, is oredicted under slimeball?

Kubuxu:
From TiC blue slime ball.

Silwing:
I also think MineFactory Reloaded's pink slimeballs are. And maybe TiC coagulated blood.

BigXplosion:
IC2 rubber and some other things that are sticky/glue related are registered as a slimeball I think

Sandriell:
The original recipe as added is bugged as well. It is looking for slimeballs with a damage value of "2", which of course does not exist.

Kubuxu:
Oh I screw up with a push. Fixing now.

asiekierka:
After thought, I decided I don't really trust mod ore dictionaries enough to allow for this. However, support for TCon/etc slimeballs will go into BuildCraftCompat.

Kubuxu:
If we tell the community that slimeballs are alternative recipe for pipe sealant they will expect all slime balls to work. Also current recipe is not working.

PR #2559, opened 2015-03-16
Lots of Changes Overall

Max-Shen0:
Added Bytecode disassembly (eg: Jasmin/Krokatau) and raw Binary modifications (such as bytecode injection).
The user is free to play modified versions of the mod, and **won't be responsible if they use an illegally modified version** (the distributor still is).
The user may not modify the license.
Also changed up the Distribution section, notably you can not distribute binaries which have been modified in the raw, since there are no corresponding sources to open (use patch files instead).
Less ambiguity, small grammatical fixes.
**"mod loading framework" is still ambiguous; not all mods even use one.** I added FMP just in case.
"this" or most recent versions of the license can now be used instead of "present" (that must have been an error, right?)
Read diffs carefully.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,23 +1,36 @@
 Minecraft Mod Public License
 ============================
 
-Version 1.0.1
+Version 1.0.2
 
 0. Definitions
 --------------
 
-Minecraft: Denotes a copy of the Minecraft game licensed by Mojang AB
+MMPL: This license.
+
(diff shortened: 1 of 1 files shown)


asiekierka:
We are not allowed to change the license without the permission of all its contributors. And if we do, we might as well write something not Minecraft-specific.

PR #2580, opened 2015-03-22
UniversalRed-patch1

UniversalRed:
I did this so that I could be able to access the BC pipe recipes for
something I was doing


asiekierka:
No. Either use reflection or wait for a proper API update.

PR #2635, opened 2015-04-15
Robot fixes

hea3ven:


Kubuxu:
This last change will break many builds and increase starting costs of robots (you have to have 3 instead of 2 docking stations and 2 instead of 1 advanced gates).

Is there any programming reason for that change?

asiekierka:
Why that last change indeed? It does make some sense, but... I am not sure if it's the right thing to do.

asiekierka:
I think it might be a good idea, though. Makes a lot of sense.

Kubuxu:
IMHO it is good idea and as 7.x is still alpha it can go through. 

Can items go into outputting part of wooden pipe? If not, it would be worth consideration to disable accepting items to wooden pipes with only one inventory connected.

hea3ven:
It does make the code simpler, but it's not required. It's mainly for the same reason of the change to accepts items. It also allows a better fine tunning of builds. 7.x already breaks a lot of builds with the change to robots only picking items from the "Load/Unload"/"Work" area.

Yes, items can go into inventories through the solid side of wooden pipe, just like any item travelling in a pipe would.

I did though of making it so that the robot could only provide an autharchic pulse and catch the items travelling through any pipe. So you'd have to set up filters with emerald pipes. But I think that's going a bit too far.

PR #2714, opened 2015-05-17
Move the Wrench Rotate Block to Server-Side.

krwminer:
Otherwise it will bypass the possible canceling of the Server-Side Forge
PlayerInteract Event which is used by Protection mods like MyTown2.


asiekierka:
That'd cause perceivable lag and is not an ideal solution. Besides, it won't "bypass" it - it will just look like it bypassed it until the block updates.

PR #2724, opened 2015-05-22
Remove crafter robot

hea3ven:
Robot items with an invalid board are converted into the basic non programmed robot item. And robot entities in the world with invalid boards are converted into "zombie" robots with the white textrure that all they do is go to sleep.


hea3ven:
Also, as a side note, not sure what SpaceToad had in mind with the RedstoneBoard, but it seems as he was thinking on using the redstone boards for more than just robots. Or maybe he was just thinking too far in the future with nothing specially in mind. The point is, right now it should be fairly simple to create redstone boards of things other than EntityRobotBase, there's only a couple places where a RedstoneBoardNBT is assumed to be a RedstoneBoardRobotNBT. Just throwing the idea around.

asiekierka:
I know that much. Also, thanks.

PR #2727, opened 2015-05-24
Change references to capacity to getCapacity()

da3dsoul:
for ease of overriding.


asiekierka:
Why?

da3dsoul:
ABO has reinforced pipes with a higher capacity and flow. The current setup either needs capacity to be protected, or to reference the getCapacity method. As of now, the absolute maximum flow is 250 mB per pipe section due to the internal mechanics of PipeTransportFluids, with no easy way to increase it.

asiekierka:
The absolute maximum flow is intended as I optimized the fluid update packet protocol to send the amount as a byte rather than a short. This needs to be rethought.

da3dsoul:
I see.

PR #2738, opened 2015-06-02
Gui Widget tooltips don't draw

mezz:
The check for `isMouseOver` in `GuiBuildCraft.mouseClicked` offsets mouseX and mouseY by guiLeft and guiTop.
The check for `isMouseOver` in `GuiBuildCraft.drawScreen` should do the same thing.

This is because Widget coordinates are relative to the Gui, not the Screen.
To make the tooltip appear, you have to move your mouse way far away right now.
Cheers :bee:


asiekierka:
Just push in Forestry's GUI code already :P /s

Merged.

mezz:
Hmm it looks like this fixes one issue but there is another. Tooltips are all offset by screen left and top in `private void drawToolTips(ToolTip toolTips, int mouseX, int mouseY)`

So now the tooltip shows up when you hover over a widget, but it's displaying far to the left and up.

psxlover:
Are you talking about #2762?

PR #2740, opened 2015-06-02
Port BuildCraft:Builders to 1.8

AlexIIL:
Current status:
 - [x] Machines are fully function on the client + integrated server
 - [ ] Machines have differing textures depending on state (little green lights if they are done, little blue lights while working...
 - [ ] Machines are fully functional on the dedicated server

All other machines work (as far as I can tell) fully on both the client and the dedicated server -but this needs proper testing.

Current bugs:
 - Land and Path markers display 3D (and tiny) in your inventory, and too low in the players hand.
 - The builder currently doesn't seem to respect its direction when building anything- but this could be related to how the architect works and there could be a bug in the architect's GUI.


asiekierka:
We already did it on Thog's branch, but we said many times we are NOT interested in 1.8 yet. Also, this is 6.2.6, not 7.0.5

PR #2802, opened 2015-06-14
Fix bugs and remove unnecessary nullchecks.

gamerforEA:
Bugs detected by FindBugs.

### DIFF
--- a/common/buildcraft/builders/BlockMarker.java
+++ b/common/buildcraft/builders/BlockMarker.java
@@ -99,7 +99,7 @@ public class BlockMarker extends BlockBuildCraft {
 		}
 
 		TileEntity tile = world.getTileEntity(i, j, k);
-		if (tile instanceof TileMarker && !(tile instanceof TileConstructionMarker)) {
+		if (tile instanceof TileMarker) {
 			((TileMarker) tile).tryConnection();
 			return true;
 		}
--- a/common/buildcraft/builders/TileQuarry.java
+++ b/common/buildcraft/builders/TileQuarry.java
@@ -205,7 +205,7 @@ public class TileQuarry extends TileAbstractBuilder implements IHasWork, ISidedI
 			// In this case, since idling() does it anyway, we should return.
 			return;
 		} else if (stage == Stage.MOVING) {
-			int energyUsed = this.getBattery().useEnergy(20, (int) Math.ceil(20D + getBattery().getEnergyStored() / 10), false);
+			int energyUsed = this.getBattery().useEnergy(20, (int) Math.ceil(20D + (double) getBattery().getEnergyStored() / 10), false);
 
 			if (energyUsed >= 20) {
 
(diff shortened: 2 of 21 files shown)


common/buildcraft/builders/BlockMarker.java
```diff
@@ -99,7 +99,7 @@ public boolean onBlockActivated(World world, int i, int j, int k, EntityPlayer e
 		}
 
 		TileEntity tile = world.getTileEntity(i, j, k);
-		if (tile instanceof TileMarker && !(tile instanceof TileConstructionMarker)) {
```
AEnterprise: why whas this removed?

AEnterprise:
you are removing alot of nullchecks, they might not always seem needed but most of them are there for a reason, you can't always assume that you are getting what you asked and that the vars are filled in correctly, it doesn't hurt to have a null check to much but it will hurt if one is missing

common/buildcraft/builders/BlockMarker.java
gamerforEA: "tile instanceof TileConstructionMarker" always return false.

asiekierka:
Actually, the checks removed are very few and all redundant.

common/buildcraft/builders/TileQuarry.java
```diff
@@ -205,7 +205,7 @@ public void updateEntity() {
 			// In this case, since idling() does it anyway, we should return.
 			return;
 		} else if (stage == Stage.MOVING) {
-			int energyUsed = this.getBattery().useEnergy(20, (int) Math.ceil(20 + getBattery().getEnergyStored() / 10), false);
+			int energyUsed = this.getBattery().useEnergy(20, (int) Math.ceil(20D + (double) getBattery().getEnergyStored() / 10), false);
```
psxlover: This should probably be:
```
int energyUsed = this.getBattery().useEnergy(20, (int) Math.ceil(20D + getBattery().getEnergyStored() / 10.0), false);
```

gamerforEA: Bytecode is identical, but your code is more compact.

psxlover: Are you sure? I don't know where Java applies the casting, so I would change the number to make sure it does a double division (10D or 10.0) or use parentesis around the casting, but if the casting applies before the division I guess it's fine like this.

gamerforEA: Casting to double, 10.0 and 10D in bytecode is identical. I check it in BytecodeOutline.

asiekierka: @psxlover - if either of the values is a double it does a double division. The Java compiler isn't that stupi.

psxlover: > @psxlover - if either of the values is a double it does a double division. The Java compiler isn't that stupi.

Yeah, I wasn't talking about that, I just wasn't sure whether the casting applied before the division or after.

PR #2814, opened 2015-06-17
Make mining robots work with gregtech ores

zmaster587:
Modified the ore check to check against what is actually dropped by the ore rather than the block itself.  This should fix almost any tile entity based ore.

Considerations:
getDrops() can be slightly slower than ItemStack(block, 1, meta) though in most cases should be unnoticeable.

### DIFF
--- a/common/buildcraft/core/properties/WorldPropertyIsOre.java
+++ b/common/buildcraft/core/properties/WorldPropertyIsOre.java
@@ -10,12 +10,14 @@ package buildcraft.core.properties;
 
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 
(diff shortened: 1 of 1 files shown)


hea3ven:
I can't check right now, but doesn't this makes blocks like redstone ore not match?

zmaster587:
Did not think about that but I'll test that and that should be an easy fix.

zmaster587:
I added the block to the list for checking as well in the second commit.  It should fix the redstone/lapis ore issue.  I also now have itemsDropped it's own list on the off-chance getDrops returns some static list for whatever reason so adding the block doesn't modify it.

asiekierka:
@hea3ven could you test this?

PR #2872, opened 2015-07-07
Change keywords so that pipes can override certain gate functions.

multiplemonomials:
When I was trying to make my Teleport Pipes transmit pipe wire signals, I realized that I had to override `updateSignalStateForColor()` and `readNearbyPipesSignal()`, set `internalUpdateScheduled`, and invoke `receiveSignal()` from a pipe subclass.  Right now, I'm using reflection to do this, but it would be a lot simpler if I could just access these fields directly.  It would just require a few changed keywords, so I'd appreciate it if you accepted this PR.

It's not just my mod, either. I think that there are several other cases in which someone would want to change these functions.

[Here](https://github.com/tcooc/AdditionalPipesBC/blob/master/src/main/java/buildcraft/additionalpipes/api/PipeTeleport.java) is the file where I use these in my mod

### DIFF
--- a/common/buildcraft/transport/Pipe.java
+++ b/common/buildcraft/transport/Pipe.java
@@ -224,7 +224,7 @@ public abstract class Pipe<T extends PipeTransport> implements IDropControlInven
 		initialized = true;
 	}
 
-	private void readNearbyPipesSignal(PipeWire color) {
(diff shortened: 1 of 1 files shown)


asiekierka:
This is bad. I plan to introduce a proper Pipe Wire API soon.

PR #2900, opened 2015-07-20
Fix filters on another pipe making items prioritize sides

Kubuxu:
When item decides where to go it has take into account two filers. Exit filter on current pipe and input filter on pipe it would go.
Exit filter should give priority for this side but input filter that is on a pipe next to current should only work as filters (no priority)
Current behaviour is really counter-intuitive when there are filters on current pipe and pipe next to it.
This change allows for chaining of golden-iron pipes with filters used to make item prioritize one side.

This also means that uncoloured items will be able to enter through input filters. The exit filter behaviour is still the same.

This makes exit filters work as priority and strict filter (no uncoloured) and input filters as refusing only wrong colours.


asiekierka:
After discussion decided to make even less sense.

PR #2919, opened 2015-07-28
fix stripes pipes placing extra blocks, and the behaviour of the stri…

hea3ven:
…pes robot placing blocks

@asiekierka not sure why you added the moveBackwards when placing blocks, but this is messing when stripes robots place blocks. This seems to be working from all directions without problem, but review it just in case I'm missing something.


asiekierka:
stripes pipes placed extra blocks due to backwards compat. I am still of the opinion that stripes robots need to die a slow death.

PR #2943, opened 2015-08-03
Architect table render distance fix + other tweaks

### LINKED ISSUE #2921: Architect Table's defined area box not visible >64 blocks
Buildcraft Version 7.0.17

Landmarks, the Builder and the Quarry show the yellow/black outline when >64 blocks away but the Architect Table does not show the blue/black outline >64 blocks away.

ganymedes01:
-Removed unused field on BlockArchitect
-Architect table area box not being visible from far away. It now uses the same render distance as the Builder. Closes #2921
-Use FluidStack sensitive version of Fluid.getLuminosity()


ganymedes01:
The last commit inspired me to create this PR, and since I was making one, I decided to check if there were any pending fixes that were easy enough to do, and I found one :3

asiekierka:
Yay!

PR #2946, opened 2015-08-04
Did a bit of a clean up

ganymedes01:
Removed a bunch of warnings by mostly removing unused fields and methods. Also improved some generics that were left undefined.

Not sure if any of the code I deleted was just left there for a reason, I just deleted anything that wasn't being used, so BuildCraft should still work exactly the same. If I deleted anything I shouldn't just let me know and I'll amend it.

### DIFF
--- a/common/buildcraft/builders/EntityMechanicalArm.java
+++ b/common/buildcraft/builders/EntityMechanicalArm.java
@@ -25,8 +25,6 @@ public class EntityMechanicalArm extends Entity {
 	private double yRoot;
 	private double zRoot;
 
-	private boolean inProgressionXZ = false;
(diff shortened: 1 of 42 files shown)


asiekierka:
The API and ChunkProperty changes need to be reverted, other than that it seems fine

ganymedes01:
Done :3
Should I add suppress warning annotations? I noticed they are used in some places but in others they aren't. I could add them if you'd like.

asiekierka:
Oh! I missed the SimplexNoise one. I think that change should be reverted as well and the class moved to core/lib to use in future BuildCraft things

ganymedes01:
Done:3

Issue #2957, opened 2015-08-11
Builder able to turn Forestry normal wood variant into the Forestry Fireproof wood variant


knexer:
Copied from https://github.com/BuildCraft/BuildcraftCompat/issues/38

BC 7.0.20
BCC 7.0.11
Forestry 1.7.10-3.6.3.20

If you have a Blueprint which contains Forestry Fireproof wood, and if you place the normal variant of that wood in the builder in a slot before the fireproof wood. Tthe builder will take the normal wood and will place down fireproof wood and not use up the fireproof wood which was in the builder.

AEnterprise:
see my comment on the other issue, BCC bug

knexer:
It's not a BCC bug; like I said in the other issue, removing Forestry compat from BCC doesn't affect the issue.  Forestry logs aren't even ITileEntityProviders (in this version of Forestry at least...), their schematic handler is being set in HeuristicBlockDetection.java to the default for non-TE blocks.

The issue is in the disagreement between the Builder's checking that requirements for the build are met and actually consuming the requirements; see this PR that fixes the issue:
https://github.com/BuildCraft/BuildCraft/pull/2958

AEnterprise:
hmm, that change might indeed fix it but won't that cause the builder to not detect other things?

knexer:
I *think* (new to this codebase) that any behavior differences will be bugs of this form.

knexer:
Right now the builder will use materials to build that don't otherwise count towards the build's requirements.  I can't see a good reason for those things to be inconsistent.  Ideally some of that logic should be deduplicated but I wanted to submit a minimal PR

PR #2958, opened 2015-08-11
Fix builder inconsistently respecting OreDict

knexer:
BptBuilderBlueprint.checkRequirements expects an exact match, but
BptBuilderBlueprint.useRequirements will accept ore-dictionary (and
other?) equivalent items.


knexer:
Fixes https://github.com/BuildCraft/BuildCraft/issues/2957

However I don't know what else this might break, especially if the Builder was set up this way intentionally.  Might be this is the wrong fix.

asiekierka:
That is a bug! A pretty big one, too.

PR #2970, opened 2015-08-14
Kinesis Pipes throughput can be changed via config

Milkyway3640:
This allows the max RF/t of a kinesis pipe to be changed via a config option. Changes to the config are also applied to the Iron pipe and Creative engine.


iTitus:
Did you actually tested that?
Maybe the PowerMode class compiles before Mc loads the config file.
Besides from that AFAIK @asiekierka does not like to have too many config options that can be used to break balance (eg. make the power throughput Integer.MAX_VALUE or something like that).

asiekierka:
BuildCraft 7 development is temporaily frozen and if I were to do it I would have done it in a less hardcoded manner that also lets you change capacities of addon pipes. Sorry.

PR #2999, opened 2015-08-31
Add a blacklist for the quarry and the mining well to stop them from mining certain blocks

### LINKED ISSUE #2998: quarry breaking blocks it shouldn't break (maybe blacklist)
reported on the forums:
http://www.mod-buildcraft.com/forums/showthread.php?tid=1810&pid=17308#pid17308

quarry seems to break protected blocks, a blacklist would be an easy solution

AlexIIL:
This fixes #2998 .


AlexIIL:
Unfortunately I messed up my pull tree (kinda, its fixable but its easier for me if you pull this first) so I won't be making any pull requests for bugfixes until this is pulled @asiekierka (I'm going to try and go through the current issue list for 7.0.21 and 7.1.x)

Kubuxu:
Isn't quarry using Fake Players?
It shouldn't be able to break those blocks.

Either way it should go to Compat if reason for it is to fix #2998.

AlexIIL:
The quarry DOES  use fake players, but it fires a forge event BreakBlock with the fake player. However the mods that I have seen do not actually listen to this event at all to test if the block can be broken (so I think its a bug report for those specific mods, but someone who makes a mod like that could prove me wrong).

Compat is (I assume) for mod-specific code that requires the actual mod to add compatibility for in the dev environment, however this code does not require any specific mods. Compat could however populate the list with sensible defaults or properly implement mod specific ignoring mechanics depending on the block in question however.

psxlover:
You can work on a branch so you wont mess with this code. You can make a pull request from any branch you create, just create on from the commit before your changes.

AlexIIL:
Thanks @psxlover but I was hoping to get this sorted quickly- looks like that isn't the case however so I will need to sort it out with git :)

asiekierka:
Poor idea. Blocks which are protected should either use events or <0 hardnesses. I'm not going to add special config options for things I don't intend to allow.

AlexIIL:
Good point.

PR #3003, opened 2015-09-06
More Ore Dictionary support for recipes

Adaptivity:
Chests and crafting tables were registered in Ore Dictionary to support Wood Stuff mod by ganymedes01. I've decided to make a PR for 7.0.x branch instead of 7.1.x since 7.1.x still in Beta and there are no ETA when it's out from Beta. I've tested this in-game and didn't found any issues, so it's shouldn't hurt anything. (I know very little about coding mods, so I'm not sure that I've put OreDictionary.registerOre strings into correct place but at least it works)

### DIFF
--- a/common/buildcraft/BuildCraftCore.java
+++ b/common/buildcraft/BuildCraftCore.java
@@ -364,6 +364,9 @@ public class BuildCraftCore extends BuildCraftMod {
 		FMLCommonHandler.instance().bus().register(this);
 		MinecraftForge.EVENT_BUS.register(this);
 		MinecraftForge.EVENT_BUS.register(new BlockHighlightHandler());
+
+		OreDictionary.registerOre("chestWood", Blocks.chest);
+		OreDictionary.registerOre("craftingTableWood", Blocks.crafting_table);
 	}
 
 	@Mod.EventHandler
--- a/common/buildcraft/BuildCraftTransport.java
+++ b/common/buildcraft/BuildCraftTransport.java
@@ -648,7 +648,7 @@ public class BuildCraftTransport extends BuildCraftMod {
 		
 		CoreProxy.proxy.addCraftingRecipe(new ItemStack(filteredBufferBlock, 1),
 				"wdw", "wcw", "wpw", 'w', "plankWood", 'd',
-				BuildCraftTransport.pipeItemsDiamond, 'c', Blocks.chest, 'p',
+				BuildCraftTransport.pipeItemsDiamond, 'c', "chestWood", 'p',
 				Blocks.piston);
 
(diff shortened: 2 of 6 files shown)


common/buildcraft/BuildCraftTransport.java
```diff
@@ -676,11 +676,11 @@ public void loadRecipes() {
 				String dye = ColorUtils.getOreDictionaryName(15 - i);
 				GameRegistry.addRecipe(new ShapedOreRecipe(new ItemStack(lensItem, 8, i),
 						"OSO", "SGS", "OSO",
-						'O', "ingotIron", 'S', dye, 'G', "blockGlass"
+						'O', "ingotIron", 'S', "dye", 'G', "blockGlass"
 						));
 				GameRegistry.addRecipe(new ShapedOreRecipe(new ItemStack(lensItem, 8, i + 16),
 						"OSO", "SGS", "OSO",
-						'O', Blocks.iron_bars, 'S', dye, 'G', "blockGlass"
+						'O', Blocks.iron_bars, 'S', "dye", 'G', "blockGlass"
```
AEnterprise: if you change this then get it out of the for loop as you are now registering the same recipe multiple times

common/buildcraft/BuildCraftCore.java
```diff
@@ -364,6 +364,9 @@ public void loadConfiguration(FMLPreInitializationEvent evt) {
 		FMLCommonHandler.instance().bus().register(this);
 		MinecraftForge.EVENT_BUS.register(this);
 		MinecraftForge.EVENT_BUS.register(new BlockHighlightHandler());
+
+		OreDictionary.registerOre("chestWood", Blocks.chest);
+		OreDictionary.registerOre("craftingTableWood", Blocks.crafting_table);
```
AEnterprise: maybe check if it isn't registered yet? not sure what the effect of double entries would be

Adaptivity: Double entries makes no effect. It's safe to use just OreDictionary.registerOre.

common/buildcraft/BuildCraftTransport.java
Adaptivity: Should I just remove "for (int i = 0; i < 16; i++) {String dye = ColorUtils.getOreDictionaryName(15 - i);"?

AEnterprise: on second look: revert this change, these recipes are for making differently colored lenses

Adaptivity:
I hope @asiekierka can merge two commits because Github for Windows very derpy thing and doesn't allow me to squash commits. Or I'll create another PR :)

PR #3013, opened 2015-09-12
Parameterization and info about new versions and changelog improved

tambry:
Now you're notified how to view the changelog, if there's a new update available. Also added info about a new version to '/buildcraft version', as it's said to include it in the update notification on startup.


asiekierka:
Why the parametrization everywhere? Especially in the API -- does it break backwards compat?

And why are you commiting into master? Master is the release branch - commit into 6.5.x if it's non-breaking (that's our 7.0 branch due to historical reasons) or 7.2.x if it's breaking (which will soon be renamed to 7.1.x, again, historical reasons).

tambry:
@asiekierka I couldn't find much better to do and I usually like having a low amount warnings, it's a personal perfomance. About backwards compatibility - not too sure, so I have rebased to 7.2.x and will make a new PR.

PR #3046, opened 2015-10-01
Fix compositeTank capacity calculation

Indemnity83:
Correct a logic error in the calculation of the capacity of a tank stack (compositeTank) which could cause an incorrect total capacity calculation if individual tiles in the stack have different capacities.

Line 264 was incorrectly using the wrong tank object (the current one, not the bottom one). This fixes indemnity83/irontank#6 and possibly others where the BuildCraft TileTank is extended.

### DIFF
--- a/common/buildcraft/factory/TileTank.java
+++ b/common/buildcraft/factory/TileTank.java
@@ -261,7 +261,7 @@ public class TileTank extends TileBuildCraft implements IFluidHandler {
 
 		TileTank tile = getBottomTank();
 
-		int capacity = tank.getCapacity();
+		int capacity = tile.tank.getCapacity();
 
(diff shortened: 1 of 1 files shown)


common/buildcraft/factory/TileTank.java
```diff
@@ -261,7 +261,7 @@ public FluidStack drain(ForgeDirection from, FluidStack resource, boolean doDrai
 
 		TileTank tile = getBottomTank();
 
-		int capacity = tank.getCapacity();
+		int capacity = tile.tank.getCapacity();
```
Thog: This should be after the null check, tile can be null.

asiekierka:
Due to Thog's note and the relative triviality of the fix, I'm going to do it myself in the fixed manner.

PR #3047, opened 2015-10-02
Improve tank texturing logic

Indemnity83:
The side texturing logic for the block is specifically looking for 'this' block instead of looking more generically if the block is an instance of BlockTank which means blocks extended from BlockTank do not render correctly when mixed. 

This brings the graphical portion of the BlockTank in-line with the way the TileTank actually works (checking instanceof when moving up and down a stack, not 'this').


asiekierka:
More bugfixes! Yay!

PR #3123, opened 2015-11-08
Fix #3119

### LINKED ISSUE #3119: Repair enchants don't work properly
In 7.1.12 the change log says that Mining Robots take advantage of enchants now. Should this extend to any robot that uses a tool? I have given a Tinker's Construct tool with Repair on it to a Lumberjack Robot, but the tool doesn't actually repair itself while the robot holds it.

iTitus:
CAUTION: This may cause problems with mods which do unchecked casts to EntityPlayer!
CAUTION: Some mods could also assume the item with the index 0 is always the held item!

It should work with TiC, because its casts to EntityPlayer are checked. Only the casts to EntityLivingBase are unchecked, but luckily the robot is an instance of that ;)

So this is just an idea.

### DIFF
--- a/common/buildcraft/robotics/EntityRobot.java
+++ b/common/buildcraft/robotics/EntityRobot.java
@@ -11,9 +11,11 @@ package buildcraft.robotics;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Set;
 import java.util.WeakHashMap;
 
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Sets;
 import com.mojang.authlib.GameProfile;
 import com.mojang.authlib.properties.Property;
 import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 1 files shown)


asiekierka:
Do the carried ItemStacks need to be ticked? Remember that they're not ticked in inventories.

iTitus:
Ok, that sounds reasonable

iTitus:
For the carried stacks: I thought these stacks could be seen as hotbar stacks. Some mods do special things if there is an item in the hotbar.

Kubuxu:
I wouldn't just ignore the exception, it can have significant performance penalty (throwing exception equals creating stack dump and so on), catch it, show message and blacklist this item for this runtime.

Just mine IMHO.

iTitus:
What do you mean with blacklisting?

asiekierka:
Make an internal HashSet<Item> which stores items which crashed, and if an
item crashes add it to that HashSet. Check if the HashSet contains a
blacklisted item before trying to onUpdate.

Also, they can be seen as hotbar stacks, yes. That works.

2015-11-08 19:36 GMT+01:00 iTitus <notifications@github.com>:

iTitus:
ok, implemented it

AEnterprise:
maybe add a way of seeing that list of items (a subcommand maybe) so people can check if an item will tick or not?

asiekierka:
The stack trace should be enough.

PR #3130, opened 2015-11-14
Fix #3128

### LINKED ISSUE #3128: Fluid Tanks displaying liquids as white (even with their color set)
Hello, one of the developers from [Growthcraft](https://github.com/Alatyami/Growthcraft-1.7) here, I'd like to report a bug regarding fluid tanks.

See, we were building against 7.0.9 for awhile, until the "a new version is available" message started nagging me, so I updated to 7.1.13, sadly.

Fluid tanks stopped displaying our fluids correctly.
Due to the number of fluids we have, we used 1 texture (a white one) along with the Fluid#getColor to denote different fluids, sadly after upgrading our fluids are rendered as their base texture without color.

[Expected (From 7.0.9)](https://dl.dropboxusercontent.com/u/72618186/screenshots/minecraft/growthcraft/tanks/2015-11-14_11.02.18.png)
[Actual (Now 7.1.13)](https://dl.dropboxusercontent.com/u/72618186/screenshots/minecraft/growthcraft/tanks/2015-11-14_11.06.41.png)
  
In addition, I've noticed that the tanks use the downcased fluidName (Possibly `Fluid#getName()`) instead of unlocalized name (`Fluid#getUnlocalizedName()`)

iTitus:
Use the Fluid's method for getting the color multiplier. I don't think that this has a serious performance impact, except when mods do heavy things in that method.

### DIFF
--- a/common/buildcraft/factory/render/RenderTank.java
+++ b/common/buildcraft/factory/render/RenderTank.java
@@ -30,10 +30,13 @@ public class RenderTank extends TileEntitySpecialRenderer {
 		TileTank tank = (TileTank) CoreProxy.proxy.getServerTile(tileentity);
 
 		FluidStack liquid = tank.tank.getFluid();
-		int color = tank.tank.colorRenderCache;
-		if (liquid == null || liquid.amount <= 0) {
+		if (liquid == null || liquid.getFluid() == null || liquid.amount <= 0) {
 			return;
 		}
+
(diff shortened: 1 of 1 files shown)


asiekierka:
You're doing it wrong. getServerTile() is an optional and not-always-working way to get the server-side tile on the client-side in single player mode to make things animate smoother. You need to support *both* methods.

iTitus:
The tank is not really "animated". I tested it, on SP and MP. There were no issues - everything run smooth.

PR #3135, opened 2015-11-20
Allow builders to place floating tripwires

ProgrammerHero:
The builder was using BlockFloored for the tripwire(string), but tripwires can be placed in the air without a supporting block - this allows this behaviour.


Kubuxu:
They can be placed above air or full block, you can't place them above ie. stairs.

ProgrammerHero:
According to vanilla 1.7.10 you *can place* them anywhere. Its only if you try to connect the wire with hooks above upside-down stairs or slabs (and possibly a couple of extra cases) they will pop off.

PR #3151, opened 2015-12-08
Merge from my development

AlexIIL:
Still very buggy, but hopefully most are small bugs that can easily be squished.


AEnterprise:
now those are ALOT of commits :P nice work on all of that

PR #3200, opened 2016-02-06
middleReached never true on Item Transport Pipes

MoreThanHidden:
Fixes: Item pipes don't work properly AlexIIL/Buildcraft_1.8_Issues#55

Can't see the need to multiply by item speed, as is already comparing the base container and travelling item positions.


AlexIIL:
Thanks! I'm just checking it...

PR #3201, opened 2016-02-07
Prevent Crash and error logs from coreState.pipeId

### LINKED ISSUE #3199: [7.2.0-pre9][Cross-mod] java.lang.IllegalArgumentException: Empty string not allowed
Placing a quartz kinesis pipe with Actually Addons' Placer causes crash: http://pastebin.com/fdFUBVXu

The idea was to use a combination of breaker and placer (both from Actually Addons) to switch quarry on or off based on how full the chest it output to is.

MoreThanHidden:
This fixes #3199 and AlexIIL/Buildcraft_1.8_Issues#109 

Prevents errors and also allowed removal of the invalid/invisible pipes from my test world, although doesn't explain the cause, added todo.


AlexIIL:
I would prefer to crash right when we know something is wrong (so if we tick and we have an invalid ID, or if we load and we have an invalid ID) so we can find out what caused it, rather than just ignoring it and deleting pipes.

MoreThanHidden:
Yeah, No worries, after I had made the commit thought this would be the case.

PR #3204, opened 2016-02-10
Item Transport Pipe improved aesthetics

MoreThanHidden:
Related to: AlexIIL/Buildcraft_1.8_Issues#55 and #3200
Improved item movement.

### DIFF
--- a/common/buildcraft/transport/PipeTransportItems.java
+++ b/common/buildcraft/transport/PipeTransportItems.java
@@ -368,12 +368,16 @@ public class PipeTransportItems extends PipeTransport implements IDebuggable {
     }
 
     protected boolean middleReached(TravelingItem item) {
-        //float middleLimit = item.getSpeed() * 1.01F;
-        return Utils.convertMiddle(container.getPos()).subtract(item.pos).lengthVector() < 0.3;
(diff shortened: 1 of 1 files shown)


asiekierka:
Wait. Why did AlexIIL touch the item movement logic to begin with?

AlexIIL:
@asiekierka when I was porting it to begin with. (I changed a bunch of stuff to Vec3 and broke everything)

AlexIIL:
Anyway, this only half works- now items travelling vertically don't  work.

MoreThanHidden:
> Anyway, this only half works- now items travelling vertically don't work.

Not sure why that is they are working in my tests...

AlexIIL:
For some reason items are vanishing for a small section. I have no idea what causes that but I don't think that it is this.

AlexIIL:
I added something like this a while ago, so now the item vanishing problem is smaller than before.

MoreThanHidden:
Yeah sorry haven't had a chance to look into thoroughly, comparing with 7.1 a couple of the comparisons are greater-than rather than less-than ext. (currently working on my own "broken aspect ratio" issues for another mod :P) but I'll take a look when I can, if you haven't already fixed it.

PR #3210, opened 2016-02-21
Create TransportExeption.java

n00b7:
to prevent errors from fluid/power/item pipes


AlexIIL:
This doesn't actually do anything useful though, can you elaborate on where this error should be thrown from and why the default Java exception classes don't provide enough information.

AEnterprise:
just making the exception class is useless, i don't see how it could be usefull like @AlexIIL already asked but without any implementation this just adds an useless and unused class to the codebase

AEnterprise:
also this isn't even written correctly, it wouldn't even compile

PR #3333, opened 2016-05-08
More checks for null

miuirussia:
Need for fix Immersive Engineering and some more mod that ruins world if connect BC Pipes


AlexIIL:
If getCapability does not work with hasCapability then there is little point in having the hasCapability method as well.

miuirussia:
Here more information: https://github.com/BluSunrize/ImmersiveEngineering/issues/1106

asiekierka:
If hasCapability returns true, getCapability must return a non-null object. We're not going to support bugs in other mods, sorry. Report to Blu.

AEnterprise:
99% of the mods will be returning the correct value, the only ones that don't are probably multiblocks or other dynamic things, if the reporter can tell what block the pipe was being attached to is enough to trace it and get it fixed

Issue #3304, opened 2016-04-14
Wrong description for Robotics module


rapus95:
https://github.com/BuildCraft/BuildCraft/blob/7.2.x/buildcraft_resources/mcmod.info#L109

states that the robotics module is the transport component

PR #3348, opened 2016-05-24
Adding some neat lighting effects for the laser (some sort of glowing, virtually)

rapus95:
@AlexIIL what you think of it? (Hoping that it'll do it into 7.2.7 :D)

Current State:
default old
Appearence after this pull request:
default enhancement

And if you've installed shaders the difference is even bigger:
shaders old
to
shaders enhancement

### DIFF
--- a/common/buildcraft/silicon/render/RenderLaserTile.java
+++ b/common/buildcraft/silicon/render/RenderLaserTile.java
@@ -25,8 +25,11 @@ public class RenderLaserTile extends TileEntitySpecialRenderer<TileLaser> {
             GL11.glTranslated(-laser.getPos().getX(), -laser.getPos().getY(), -laser.getPos().getZ());
 
             GL11.glPushMatrix();
+            GlStateManager.disableLighting();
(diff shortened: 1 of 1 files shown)


asiekierka:
Was in 1.7.10, I believe - so this brings it back. Heh.

AlexIIL:
10/10 would pull again :D

rapus95:
Btw I again only was the one doing the github stuff :D @XOR19 did the work

AlexIIL:
Can both you and @XOR19 take a look at #3304 then? I just noticed that nether of you have agreed to a re-license, and I don't really want to find out in the future that either of you don't agree AFTER it has been re-licensed :)

rapus95:
@AlexIIL I'm representing the opinion that i don't code for any personal rights but for better playing experience (especially when playing myself :D) thus there wouldn't have been a problem with getting my permission even at any time in the future :)

AlexIIL:
@rapus95 @XOR19 thanks :)

PR #3350, opened 2016-05-24
fixes #3311

### LINKED ISSUE #3311: robots have problems with torches and partial/half blocks 7.2.5
the picker robot has huge problems with items falling into half blocks (like onto chests  or pipes etc) aswell as falling into a space that is occupied by a torch

Edit: added screenshots

XOR19:
a) blocks without boundingbox can now be in a path
b) Robots (picker) can now pickup items one block away (if block has a bounding
box)
c) robots shouldn't be stuck due to path changes anymore (hopefully recalculates path)


AlexIIL:
Seems to work very nicely :)

Issue #3226, opened 2016-03-13
Weird block placement on engine


LuigiHutch:
BC 7.2.0

Blocks place weirdly on engines, See Gfycat:
https://gfycat.com/ExemplaryUnhappyConey

AlexIIL:
I have no idea what causes this :(

PR #3351, opened 2016-05-25
fixes #3320

### LINKED ISSUE #3320: 7.2.5 Render error in the list GUI
When opening the gui of the list item every enchanted item in the upper row of the player inventory has a render bug exactly then when you don't mouse-over any of the buildcraft slots or the upper row of the player inv (while all other slots are rendered fine in either case) (in the pictures it's the pickaxe)

(Mouse not over one of the two rows of slots)

(Mouse over the 4th slot of the 2nd row)

XOR19:
enchantment animation isn't hiding the item anymore in the GuiList, fixed through the GlStateManager.disableBlend(), but don't know why that's needed.
Also moved second row of slots one pixel down, cause it didn't align with the
background

### DIFF
--- a/buildcraft_resources/assets/buildcraftfactory/blockstates/floodGateBlock.json
+++ b/buildcraft_resources/assets/buildcraftfactory/blockstates/floodGateBlock.json
@@ -1,5 +1,15 @@
 {
+    "forge_marker":1,
     "variants": {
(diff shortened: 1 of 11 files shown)


XOR19:
and also fixes #3226 which was caused due to that you took the last hitbox in the list, not the nearest

PR #3354, opened 2016-05-25
map location update (as planned in the forum)

XOR19:
only blank map locations can be written by right click, used map
locations can be cleared with sneak right click, left click passes through
item to block.
also paths should work again, it seems like it didn't before caused by that the path list contained int array tags but when read expected tag compounds.

### DIFF
--- a/common/buildcraft/core/ItemMapLocation.java
+++ b/common/buildcraft/core/ItemMapLocation.java
@@ -14,10 +14,10 @@ import net.minecraft.nbt.NBTTagList;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.BlockPos;
 import net.minecraft.util.EnumFacing;
+import net.minecraft.util.MathHelper;
 import net.minecraft.util.MovingObjectPosition;
+import net.minecraft.util.Vec3;
 import net.minecraft.world.World;
-
-import net.minecraftforge.common.util.Constants;
 
 import buildcraft.api.core.IAreaProvider;
 import buildcraft.api.core.IBox;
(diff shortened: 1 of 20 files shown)


XOR19:
fix for maybe some render issues like #3355, caused by wrong box-encompass calculation

common/buildcraft/core/ItemMapLocation.java
```diff
@@ -93,64 +96,102 @@ public void addInformation(ItemStack stack, EntityPlayer player, @SuppressWarnin
...
+        float f5 = MathHelper.sin(-f * 0.017453292F);
+        float f6 = f3 * f4;
+        float f7 = f2 * f4;
+        double d3 = 5.0D;
+        if (playerIn instanceof net.minecraft.entity.player.EntityPlayerMP) {
+            d3 = ((net.minecraft.entity.player.EntityPlayerMP) playerIn).theItemInWorldManager.getBlockReachDistance();
         }
+        Vec3 vec31 = vec3.addVector((double) f6 * d3, (double) f5 * d3, (double) f7 * d3);
+        return worldIn.rayTraceBlocks(vec3, vec31, false, false, false);
+    }
```
AlexIIL: Why is this necessary? Is this copied from somewhere in minecraft or did you just write all of this?

XOR19: cpy from Item.getMovingObjectPositionFromPlayer() but these the problem that i don't get the last three paremeters of the call to worldIn.rayTraceBlock to false, only false true false or true false false what would cause blocking on water or the landmaks get not accepted

AlexIIL:
I thought I fixed the ````Box.extendToEncompass```` being wrong, but it turns out I only fixed it in the 8.0.x branch. I guess I'll have to re-look at all my fixes from there to see what else needs back-porting.

XOR19:
but in the 8.0.x branch Box.extendToEncompass(IBox toBeContained) you should look to take the min and max and not directly set it. like this its more a Box.setTo(IBox toBeSetTo)

AlexIIL:
Looks like I missed MORE bugs!

AlexIIL:
Ok, so I did this in https://github.com/BuildCraft/BuildCraft/commit/007ee969dba48b7ed1d0aeb5742467c65f98c4f1 but without changing as much stuff. I recommend using a different branch for each PR if you want to contribute more (so that minor changes can be merged back into your branches more simply.)

XOR19:
ok, but @AlexIIL you forgot to fix the Box.extendToEncompass(IBox) in the 7.2.x branch

AlexIIL:
Hopefully that's it now :)

Issue #3004, opened 2015-09-09
Change BuildCraft's license


asiekierka:
Long overdue. The MMPLv1 has a few nasty holes and is incompatible with a lot of other FOSS licenses due to at least the "legal copy of Minecraft" clause.

Since one of our primary contributors' request was to choose a "copyleft license with minimum lawyerspeak", I vote for the **[Mozilla Public License 2.0](https://www.mozilla.org/en-US/MPL/2.0/)**, used by Mozilla for their own projects as well as many other developers who are not okay with the GPL's high level of protection.

I'm going to start calling people as I figure out what parts of the code exactly need re-licensing, but let me call some of the primary and recent contributors off memory:

SpaceToad (will contact him separately)
@CovertJaguar 
@SirSengir 
@Krapht 
@hea3ven 
@cpw 
@ganymedes01 
@Flow86 
@tambry 
@AartBluestoke 

I need your permissions first.

Krapht:
This again? Fine...

I permit the license of my contributions to the BuildCraft project to be changed to the Mozilla Public License, v. 2.0

ganymedes01:
So do I.
I have no problem with the license change.

asiekierka:
@Krapht - my original intention was MIT, but one of our contributors refused; it took me a while to pick a license that fit their requests.

hea3ven:
I agree with the license change.

pahimar:
I don't care at all, I honestly can't remember what code I've even contributed. I've long ago stopped following this project.

asiekierka:
@pahimar - the threaded BC version checker, off the top of my head. Still - even a few lines of code matter and I'd rather have the legal clear as we never had CLAs so you own your code!

Itaros:
I permit the license of my contribution to the BuildCraft project to be changed to the Mozilla Public License, v. 2.0 or MIT License

davboecki:
I permit the license of my contributions to the BuildCraft project to be changed to the Mozilla Public License, v. 2.0.

smbarbour:
I know I have contributed in the past, so "I permit the license of my contributions to the BuildCraft project to be changed to any license that the project managers see fit"

ljfa-ag:
I only contributed some tiny changes iirc, but just in case you need it you have my blessing as well.

anti344:
Lol, couple of my tiny fixes now need my permission =P
(i just remembered, i fixed super-annoying laser and engine selection boxes :D)

[...]

PR #3377, opened 2016-06-28
8.0.x

afdw:
3D book, new forge version and auto workbenches

### DIFF
--- a/common/buildcraft/api/bpt/BlueprintAPI.java
+++ b/common/buildcraft/api/bpt/BlueprintAPI.java
@@ -12,7 +12,7 @@ import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.StringUtils;
 
-//import buildcraft.api.IUniqueReader;
+import buildcraft.api.IUniqueReader;
 
 public class BlueprintAPI {
     private static final Map<ResourceLocation, SchematicFactoryWorldBlock> worldBlockFactories = new HashMap<>();
@@ -94,8 +94,7 @@ public class BlueprintAPI {
         NBTTagCompound data = nbt.getCompoundTag("data");
         IBptReader<IBptAction> reader = getActionDeserializer(new ResourceLocation(id));
         if (reader != null) {
-//            return reader.deserialize(data); // FIXME: Anton
-            return null; // FIXME: Anton
+            return reader.deserialize(data);
         } else {
             return null;
         }
--- a/common/buildcraft/api/bpt/BlueprintAPI.java
+++ b/common/buildcraft/api/bpt/BlueprintAPI.java
@@ -12,7 +12,7 @@ import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.StringUtils;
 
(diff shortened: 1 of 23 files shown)


common/buildcraft/api/bpt/BlueprintAPI.java
```diff
@@ -94,7 +94,8 @@ public static IBptAction deserializeAction(NBTTagCompound nbt) {
         NBTTagCompound data = nbt.getCompoundTag("data");
         IBptReader<IBptAction> reader = getActionDeserializer(new ResourceLocation(id));
         if (reader != null) {
-            return reader.deserialize(data);
+//            return reader.deserialize(data); // FIXME: Anton
+            return null; // FIXME: Anton
```
AlexIIL: I've just fixed this in a commit.

AlexIIL:
This is interesting, I didn't think I would get help for this!

Although a few things have (just) changed because of a commit I just made.

Because BC is undergoing a relicense you would have to agree to it (here)[https://github.com/BuildCraft/BuildCraft/issues/3004] before I can accept this. Also all of the "buildcraft.api" packages are going to be relicensed to MIT and put in the official API repository when its settled down, so you would also need to agree to them moving and being under MIT rather than MPL2.

AEnterprise:
@AlexIIL due to the new BC code license i sugest adding a CLA to BC, github has integration for it that can do automatic checking and indicating if the contributors in a PR signed or not

common/buildcraft/api/bpt/BlueprintAPI.java
```diff
@@ -12,7 +12,7 @@
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.StringUtils;
 
-import buildcraft.api.IUniqueReader;
+//import buildcraft.api.IUniqueReader;
```
AEnterprise: i would advise agains commenting out imports, it's easy to have your IDE auto-re-import but the comment will stay there being useless

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sat May 21 15:16:57 BST 2016
+#Tue Jun 28 17:10:16 MSK 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-all.zip
```
AEnterprise: do we want to use the "all" version? the bin should be enough and it's just a larger download that isn't needed i think

AlexIIL:
Seems good, thanks :)

AlexIIL:
So I've just tested the new Guide Opening animation you made. Its awesome :D

LuigiHutch:
Gif plz :)

afdw:
@AlexIIL Is there any place (chat, mail, etc) we can talk?

AlexIIL:
@afdw The IRC channel "buildcraft" (espernet) is probably the best place, however you can always PM me on there if it has to be private.

PR #3379, opened 2016-07-01
Start porting Mining Well

afdw:


AEnterprise:
can you do PR's with more descriptive titles? some more info in the commit messages of what was ported/fixed/... would be good as well for making changelogs and such

i also notice you are using spaces instead of tabs but the rest of the repo uses tabs, could you change your formatter to use the BC one? otherwise some files are gona become a mess in the end

AlexIIL:
@afdw This looks good, I'm not too sure about ````IBlockWithFacing```` however enough blocks need it so its probably fine. Hopefully I won't forget to re-add power consumption in :)

@AEnterprise The entire codebase was reformatted to spaces when I merged my port in so if you could revert that commit it would be good :)

afdw:
@AlexIIL OK, I reverted it.

AEnterprise:
@AlexIIL ah i didn't know that, been a while since i wrote code for BC

AlexIIL:
@afdw if you pull https://github.com/afdw/BuildCraft/pull/1 then I'll pull this :) (I should have done that commit AFTER pulling this, sorry)

afdw:
OK

AlexIIL:
Thanks :)

PR #3380, opened 2016-07-01
Pump start and auto workbenches

afdw:


common/buildcraft/factory/tile/TileAutoWorkbenchBase.java
```diff
@@ -71,9 +75,73 @@ protected void disableBindings() {
...
+                        if(StackUtil.canMerge(resultStack, currentRecipe.getRecipeOutput())) {
+                            resultStack.stackSize += currentRecipe.getRecipeOutput().stackSize;
+                        }
+                    } else {
+                        resultStack = currentRecipe.getRecipeOutput().copy();
+                    }
+                }
+            }
+            invResult.setStackInSlot(0, resultStack);
+        }
```
AlexIIL: Unfortunately its a lot more complicated than this - you have to call ````IRecipe.getRemainingItems```` and ````IRecipe.getCraftingResult```` to properly craft the recipe as different custom recipe handlers do different things. Which is where all of the bindings come in.

afdw: And... How can I use getRemainingItems here? In this code I get resources that is in craft matrix. They are already checked with CraftingUtils.findMatchingRecipe. After that I try to remove resources from materials inventory and add output given with getRecipeOutput

afdw: OK, now I use getCraftingResult, but I still not understand, how to use getRemainingItems.

PR #3382, opened 2016-07-01
Pumps and leds

afdw:


common/buildcraft/factory/client/render/RenderLeds.java
```diff
@@ -0,0 +1,70 @@
...
+                    bindTexture(new ResourceLocation("buildcraftcore", "textures/blocks/led_green_off.png"));
+                } else {
+                    bindTexture(new ResourceLocation("buildcraftcore", "textures/blocks/led_green_on.png"));
+                }
+                new ModelRenderer(this).addBox(tile.getX(), tile.getY(), -0.1F, 1, 1, 1).render(scale);
+                bindTexture(new ResourceLocation("buildcraftcore", "textures/blocks/led_red_" + MathUtils.clamp(tile.getPowerLevel(), 0, 3) + "_3.png"));
+                new ModelRenderer(this).addBox(tile.getX() - 2, tile.getY(), -0.1F, 1, 1, 1).render(scale);
+            }
+        };
+        model.render(null, 0, 0, 0, 0, 0, 0.0625F);
```
AlexIIL: This is very inefficient. But it works, and it can always be optimised later.

common/buildcraft/factory/client/render/RenderLeds.java
```diff
@@ -0,0 +1,70 @@
...
+    public static final RenderLeds INSTANCE = new RenderLeds();
+
+    @Override
+    public void renderTileEntityAt(T tile, double x, double y, double z, float partialTicks, int destroyStage) {
+        RenderHelper.disableStandardItemLighting();
+        GL11.glColor3f(1, 1, 1);
+        int i = 61680;
+        int j = i % 65536;
+        int k = i / 65536;
+        OpenGlHelper.setLightmapTextureCoords(OpenGlHelper.lightmapTexUnit, (float)j, (float)k);
```
AlexIIL: In a shorter form this is just this:
````java
OpenGlHelper.setLightmapTextureCoords(OpenGlHelper.lightmapTexUnit, 0xF0F0, 0);
````

However the texture is rounded at 0x100 so you could just use 0xF0 instead.

Comment (afdw) [inline common/buildcraft/factory/client/render/RenderLeds.java]: Why you don't answer me in chat? Can you see me messages? I wrote about it.

File: common/buildcraft/factory/tile/ITileLed.java [hunk 3]
```diff
@@ -0,0 +1,8 @@
+package buildcraft.factory.tile;
+
+public interface ITileLed {
+    boolean isDone();
+    int getPowerLevel();
+    int getX();
+    int getY();
```
AlexIIL: These aren't very clear - how about ````getLedX```` and ````getLedY```` as then its obvious that it doesn't refer to the co-ords of the tile.

common/buildcraft/factory/client/render/RenderLeds.java
AlexIIL: Sorry.

common/buildcraft/factory/tile/ITileLed.java
```diff
@@ -0,0 +1,8 @@
+package buildcraft.factory.tile;
+
+public interface ITileLed {
+    boolean isDone();
```
AEnterprise: the led state is determined by if the machine is done or not but wouldn't isOn() or isLedOn() be more acurate?

common/buildcraft/factory/client/render/RenderLeds.java
AEnterprise: maybe add a comment where you found those magic numbers, i asume in the vanilla code so if things change you'll easely know where to find the new values

common/buildcraft/factory/client/render/RenderLeds.java
AEnterprise: probably better to make a inner class that extends modelbase and stores a ITileLed with a setter you can call just before the rendering or a custom function that also takes the tile as param to call the function. at 60FPS this is alot of junk objects for the GC to clean up

PR #3384, opened 2016-07-01
Pump

afdw:


common/buildcraft/api/mj/MjBattery.java
```diff
@@ -43,7 +43,7 @@ public void readFromBuffer(ByteBuf buffer) {
     }
 
     public void addPower(int milliJoules) {
-        this.milliJoules += milliJoules;
+        this.milliJoules = Math.min(this.milliJoules + milliJoules, capacity);
```
AlexIIL: That was explicitly done by design.

common/buildcraft/factory/client/render/RenderMiningWell.java
```diff
@@ -19,7 +19,7 @@
     private static final int COLOUR_STATUS_DONE = 0xFF_1f_10_1b; // black-ish
 
     private static final int BLOCK_LIGHT_STATUS_OFF = 0x0;
-    private static final int BLOCK_LIGHT_STATUS_TODO = 0xF;
+    private static final int BLOCK_LIGHT_STATUS_TODO = 0x1;
```
AlexIIL: Why? Do you dislike the full brightness LED or does it not work for you?

common/buildcraft/factory/tile/TileMiner.java
```diff
@@ -0,0 +1,193 @@
...
+    public void update() {
+        if (worldObj.isRemote || mode != Mode.On) {
+            return;
+        }
+        battery.tick(getWorld(), getPos());
+
+        // test with the output of a stone engine
+        battery.addPower(1000);// remove this
+
+        if (worldObj.rand.nextDouble() > 0.9) { // is this correct?
```
AlexIIL: No, that was temporary for testing. It needs to be replaced with a check to see if it should actually send an update message yet (the status of any of the LED's has changed)

common/buildcraft/factory/tile/TileMiningWell.java
```diff
@@ -1,74 +1,25 @@
...
-        if (isComplete) {
-            return;
-        }
-
-        initCurrentPos();
-
+    protected void mine() {
         IBlockState state = worldObj.getBlockState(currentPos);
-        if (BlockUtils.isUnbreakableBlock(getWorld(), currentPos)) {
+        if (BlockUtils.isUnbreakableBlock(getWorld(), currentPos) || state.getBlock() == Blocks.BEDROCK) {
```
AlexIIL: It looks like BlockUtils.isUnbreakableBlock is failing to identify bedrock as unbreakable, ideally you would not have to also check here.

common/buildcraft/factory/client/render/RenderMiningWell.java
afdw: Yes, it not working for me. Why 0xF? It's 15. Maximum is one.

common/buildcraft/factory/tile/TileMiner.java
afdw: Yes, I add this comment like TODO.

common/buildcraft/lib/fluid/FluidStorage.java
```diff
@@ -0,0 +1,147 @@
...
+                    int g = ((int) bytes[2]) & 0xFF;
+                    int b = ((int) bytes[3]) & 0xFF;
+                    fluidColors.put(getFluidFluid(), ((a & 0xff) << 24) + ((b & 0xff) << 16) + ((g & 0xff) << 8) + (r & 0xff));
+                } catch(Exception e) {
+                    e.printStackTrace();
+                }
+            }
+            return fluidColors.get(getFluidFluid());
+        }
+        return 0xFF_00_00_00;
```
AlexIIL: You can use
````java
fluid.getColor(fluidStack)
````
Instead to get the intended colour.

Comment (afdw) [inline common/buildcraft/lib/fluid/FluidStorage.java]: No, not working: return getFluidFluid().getColor(fluid);. It returns -1.

File: common/buildcraft/lib/fluid/FluidStorage.java [hunk 6]
```diff
@@ -0,0 +1,147 @@
...
+        NBTTagCompound tankData = new NBTTagCompound();
+        super.writeToNBT(tankData);
+        ByteBufUtils.writeTag(buffer, tankData);
+    }
+
+    public void readFromBuffer(ByteBuf buffer) {
+        NBTTagCompound tankData = ByteBufUtils.readTag(buffer);
+        super.readFromNBT(tankData);
+    }
+}
```
AlexIIL: You should probably use ````buildcraft.core.lib.fluids.SingleUseTank```` instead of this.

afdw: This is port of SingleUseTank.

common/buildcraft/factory/client/render/RenderMiningWell.java
AlexIIL: Woops, I was calling "lightf" rather than "lighti". That explains why it wasn't working.

afdw: But now it is workring fine.

common/buildcraft/lib/fluid/FluidStorage.java
AlexIIL: Fortunately however everyone who contributed to both "Tank" and "SingleUseTank" has agreed to re-license to MPL2 so it doesn't need to be ported :)

(Although I have only just checked that, and I hope that a lot more classes fall into this category).

common/buildcraft/lib/fluid/FluidStorage.java
AlexIIL: Fair enough.

afdw:
@AlexIIL OK, I merged with you.

AlexIIL:
Thanks, I'm now testing out how it works.

AlexIIL:
Ok, so a few things are wrong:

1. The pump should use a moving tube (rendered in a FastTESR) rather than a block when pumping.
2. The pump currently keeps on going, even if it has filled up its internal tank.
3. The pump keeps on attempting to suck up fluid, even if there is no available fluid.
4. The pump doesn't extend its tube downwards if there is a fluid a long way down (try using the 1.8.9 pump to see what I mean).
5. I *think* you have copied the old pump code over, which you can't really do as some people who contributed to the pre-8.0.x pump code haven't agreed to the relicense. Its existed for a long time, so I haven't been able to properly check.

afdw:
I now, I wrote in chat "but there are several problems". But thanks.

common/buildcraft/factory/client/render/RenderTank.java
```diff
@@ -18,43 +25,27 @@ public RenderTank() {
...
+                return true;
+            };
         }));
     }
 
     @Override
     public void renderTileEntityFast(TileTank tile, double x, double y, double z, float partialTicks, int destroyStage, VertexBuffer buffer) {
+        GL11.glEnable(GL11.GL_CULL_FACE);
         super.renderTileEntityFast(tile, x, y, z, partialTicks, destroyStage, buffer);
+//        GL11.glDisable(GL11.GL_CULL_FACE); // FIXME: Why I can't disable it here?
```
afdw: Is there other way to enable cull face?

common/buildcraft/lib/fluids/TankUtils.java
```diff
@@ -0,0 +1,24 @@
...
+                IHasTank tileToPushWithTank = (IHasTank) tileToPush;
+                int used = tileToPushWithTank.getTank().fill(tileWithTank.getTank().getFluid(), true);
+
+                if(used > 0) {
+                    tileWithTank.getTank().drain(used, true);
+                }
+            }
+        }
+    }
+}
```
AlexIIL: Is this basically horizontal transfer utils? This isn't really necessary, however it's probably required for testing the other tank blocks before we get pipes.

afdw: This is method that transfers fluid in TE to all possible TE to all sides (not only horizontal).

common/buildcraft/factory/client/render/RenderTank.java
```diff
@@ -18,43 +25,27 @@ public RenderTank() {
...
+            part.shouldSideBeRendered = side -> {
+                if(side.getAxis() == EnumFacing.Axis.Y) {
+                    if(Minecraft.getMinecraft().theWorld.getBlockState(tile.getPos().offset(side)).getBlock() instanceof BlockTank) {
+                        TileTank tileSide = (TileTank) Minecraft.getMinecraft().theWorld.getTileEntity(tile.getPos().offset(side));
+                        return tileSide.tank.isEmpty();
+                    }
+                    return true;
+                }
+                return true;
+            };
```
AlexIIL: This will probably have to be replaced with a specialised fluid renderer, as this just squishes the side texture down in odd ways. We might also need to have our own render buffer for fluid rendering, to get around the culling problem.

common/buildcraft/factory/client/render/RenderTank.java
AlexIIL: Or not use the given buffer to render.

common/buildcraft/lib/fluids/TankUtils.java
AlexIIL: Oh, so this would be used for things like pumps? That's fine then.

common/buildcraft/factory/tile/TileTank.java
```diff
@@ -0,0 +1,87 @@
...
+import net.minecraftforge.fml.relauncher.Side;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.List;
+
+public class TileTank extends TileBC_Neptune implements ITickable, IDebuggable {
+    public static final int NET_TANK = 10;
+
+    public SingleUseTank tank = new SingleUseTank("tank", 16000, this);
```
AlexIIL: You don't want to use SingleUseTank here- that only allows for a single type of fluid to be in the tank, ever. Which is fine for the pump (we don't want to start by pumping out oil and then start to pump out water) the tank just needs to accept any fluid.

afdw: And what should I use?

AlexIIL: Just normal ````Tank````

common/buildcraft/lib/fluids/TankUtils.java
```diff
@@ -4,19 +4,20 @@
...
-                int used = tileToPushWithTank.getTank().fill(tileWithTank.getTank().getFluid(), true);
+            if(tileToPush != null && tileToPush.hasCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, null)) {
+                Tank tankToPush = (Tank) tileToPush.getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, null);
+                int used = tankToPush.fill(tank.getFluid(), true);
 
                 if(used > 0) {
-                    tileWithTank.getTank().drain(used, true);
+                    tank.drain(used, true);
                 }
             }
```
AlexIIL: It would probably be better to change this so that you pass in the ````Tank```` variable rather than getting it via the capability (because at some point in the future the capability returned by pump needs to not accept any fluids externally.)

common/buildcraft/factory/client/render/RenderTank.java
```diff
@@ -0,0 +1,51 @@
...
+import org.lwjgl.opengl.GL12;
+import org.lwjgl.opengl.GL14;
+
+public class RenderTank extends RenderMultiTile<TileTank> {
+    public RenderTank() {
+        // TODO: optimize
+        parts.add(new RenderPartElement<>((tile, part) -> {
+            part.sizeY = ((double)tile.tank.getFluidAmount() / tile.tank.getCapacity()) * 16.0 / 16.0;
+            if(part.sizeY != 0) {
+                part.sizeX = part.sizeZ = 10.0 / 16.0;
```
AlexIIL: This should be a little bit smaller than the size of the tank, so ````12.0 / 16.1```` .

AlexIIL:
This should be fine. At the very least I want to do some refactors, so it's easier to pull this and refactor than the other way around.

PR #3388, opened 2016-07-05
Flood gate and chute

afdw:


common/buildcraft/lib/net/MessageUpdateTile.java
```diff
@@ -46,6 +49,7 @@ public void fromBytes(ByteBuf buf) {
     public void toBytes(ByteBuf buf) {
         PacketBuffer buffer = new PacketBuffer(buf);
         buffer.writeBlockPos(pos);
+        buffer.writeString(className);
```
AlexIIL: We do strings need to be written out? They are expensive, and relying on the correct class type seems... odd. Did you actually crash ever because a tile entity changed types too quickly?

afdw: There is a bug in network classes, so readPayload gives wrong buffer. It's for fix it for now and for debugging problem.

afdw: No, no changing types very quickly, the bug appear in static construction.

AEnterprise: i had that bug as well in BCA, instead of a string i gave every TE an a number and used that to sync, a short should be enough

afdw: Hm... We can use tile.getClass().getName().hashCode() for auto generating this number, but this is a bug, and we need to fix it. This is a hack.

AEnterprise: i know this doesn't fix the core bug, just makes the workaround lighter on the networking

common/buildcraft/factory/tile/TileChute.java
```diff
@@ -0,0 +1,212 @@
...
+            progress = 0;
+            int radius = 3;
+            BlockPos offset = new BlockPos(currentSide.getDirectionVec());
+            offset = new BlockPos(offset.getX() * radius, offset.getY() * radius, offset.getZ() * radius);
+            AxisAlignedBB aabb = new AxisAlignedBB(this.pos, this.pos).expandXyz(radius).offset(offset);
+            List<EntityItem> entityItems = worldObj.getEntitiesWithinAABB(EntityItem.class, aabb);
+            int index = 0, max = 3;
+            for(EntityItem entityItem : entityItems) {
+                ItemStack stack = entityItem.getEntityItem();
+                stack = ItemTransactorHelper.getTransactor(this, currentSide).insert(stack, false, false);
```
AlexIIL: You can just use ````inv``` here (instead of ````ItemTransactorHelper.getTransactor(this, currentSide)````) as ItemHandlerSimple implements IItemTransactor already.

common/buildcraft/factory/tile/TileChute.java
```diff
@@ -0,0 +1,212 @@
...
+        if(worldObj.isRemote) {
+            return;
+        }
+
+        battery.tick(getWorld(), getPos());
+
+        // test with the output of a stone engine
+        battery.addPower(1000); // remove this
+
+        EnumFacing currentSide = (EnumFacing) worldObj.getBlockState(pos).getProperties().get(BlockBCBase_Neptune.BLOCK_FACING_6);
```
AlexIIL: You can use ````worldObj.getBlockState(pos).getValue(BlockBCBase_Neptune.BLOCK_FACING_6)```` instead. Although I have come across times when the tile still exists but the block doesn't, so you should have a check to make sure the block is the same.

AEnterprise: TE's recieve an additional tick before being destroyed

common/buildcraft/lib/RegistryHelper.java
```diff
@@ -47,8 +48,8 @@ public static boolean registerItem(Item item) {
     public static boolean registerItem(Item item, boolean forced) {
         if (forced || isEnabled(getActiveMod(), "items", item.getRegistryName().getResourcePath())) {
             GameRegistry.register(item);
-            if (item instanceof ItemBC_Neptune) {
-                ItemBC_Neptune itemBc = (ItemBC_Neptune) item;
+            if (item instanceof IItemBuildCraft) {
+                IItemBuildCraft itemBc = (IItemBuildCraft) item;
```
afdw: Debugging isn't very fun :-(

common/buildcraft/lib/marker/MarkerSubCache.java
```diff
@@ -103,6 +103,15 @@ public void removeMarker(BlockPos pos) {
         }
     }
 
+    public void removeConnection(BlockPos pos1, BlockPos pos2) {
+        for(MarkerConnection connection : getConnections()) {
+            BlockPos[] positions = (BlockPos[]) connection.getMarkerPositions().toArray(new BlockPos[0]);
+            if(Arrays.binarySearch(positions, pos1) >= 0 && Arrays.binarySearch(positions, pos2) >= 0) {
+                deinitConnection(new HashSet<>(Arrays.asList(positions)));
```
AlexIIL: A few things:

1. Removing connections looks odd, and is generally easier to destroy the marker itself rather than the connection (as suddenly there are a lot of connection paths that looks odd). Basically "remove this please".

2. You can't perform a binary search on an unsorted array/list.

3. Just blindly using ````deinitConnection```` doesn't work.

PR #3391, opened 2016-07-11
Filtered buffer

afdw:


common/buildcraft/transport/container/ContainerFilteredBuffer.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.client.renderer.texture.TextureAtlasSprite;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.ItemStack;
+
+public class ContainerFilteredBuffer extends ContainerBCTile<TileFilteredBuffer> {
+    private static final int PLAYER_INV_START = 86;
+    private static SpriteHolderRegistry.SpriteHolder emptyFilterSlotHolder;
+
+    public static void preInit() {
+        emptyFilterSlotHolder = SpriteHolderRegistry.getHolder("buildcrafttransport:gui/empty_filtered_buffer_slot");
```
AlexIIL: You should probably move this to a client-only class as sprites don't exist on the server (so this will probably crash when you start up a dedicated server).

I think it would be better to have a dedicated class for transport sprites (as there are quite a few of them).

You can also just use
````java
public static final SpriteHolder SPRITE_THING = SpriteHolderRegistry.getHolder("location:of/thing");
````
Provided you have a method *somewhere* in that class for preInit, which is called by the client proxy.

File: common/buildcraft/transport/tile/TileFilteredBuffer.java [hunk 2]
```diff
@@ -0,0 +1,16 @@
...
+import buildcraft.lib.tile.item.ItemHandlerManager;
+import buildcraft.lib.tile.item.ItemHandlerSimple;
+
+public class TileFilteredBuffer extends TileBCInventory_Neptune {
+    public final ItemHandlerSimple invFilter;
+    public final ItemHandlerSimple invMain;
+
+    public TileFilteredBuffer() {
+        invFilter = addInventory("blueprint", 9, ItemHandlerManager.EnumAccess.NONE);;
+        invMain = addInventory("materials", 9, ItemHandlerManager.EnumAccess.INSERT, EnumPipePart.VALUES);
```
AlexIIL: The idea with the filtered buffer is that it only allows items to go into the "main" inventory if the item matches what is in the blueprint inventory. You can also extract from the filtered buffer's main inventory.

common/buildcraft/lib/gui/GuiBC8.java
```diff
@@ -73,11 +73,11 @@ protected final void drawGuiContainerBackgroundLayer(float partialTicks, int mou
 
     @Override
     protected final void drawGuiContainerForegroundLayer(int mouseX, int mouseY) {
+        drawForegroundLayer();
+
         GlStateManager.translate(-guiLeft, -guiTop, 0);
         mouse.setMousePosition(mouseX, mouseY);
 
-        drawForegroundLayer();
-
```
AlexIIL: Swapping these around breaks all other GUI's.

afdw: All other GUI's fixed.

AlexIIL: Sorry, it was changed to align drawForegroundLayer with drawBackgroundLayer so they would both start from the same co-ordinates. So can you change it back?

PR #3394, opened 2016-07-12
Zone planner

afdw:


loordgek:
@afdw very cool that your helping @AlexIIL  :)

asiekierka:
The zone planner code should be mostly fine to use.

common/buildcraft/lib/BCMessageHandler.java
```diff
@@ -53,6 +55,8 @@ public static void fmlPreInit() {
 
     public static void fmlPostInit() {
         netWrapper = NetworkRegistry.INSTANCE.newSimpleChannel("bc-lib");
+        netWrapper.registerMessage(MessageZonePlannerMapChunkRequest.Handler.class, MessageZonePlannerMapChunkRequest.class, 123, Side.SERVER); // TODO: move it form here
+        netWrapper.registerMessage(MessageZonePlannerMapChunkResponse.Handler.class, MessageZonePlannerMapChunkResponse.class, 124, Side.CLIENT);
```
AlexIIL: This should be in the BCRobotics class, in ````preInit````. It should *probably* look like this:
````java
BCMessageHandler.addMessageType(MessageZonePlannerMapChunkRequest.class, MessageZonePlannerMapChunkRequest.Handler.INSTANCE, Side.SERVER);
BCMessageHandler.addMessageType(MessageZonePlannerMapChunkResponse.class, MessageZonePlannerMapChunkResponse.Handler.INSTANCE, Side.CLIENT);
````

This does require you to change the handlers to enum types, with a single object "INSTANCE".

File: common/buildcraft/robotics/MessageZonePlannerMapChunkRequest.java [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static class Handler implements IMessageHandler<MessageZonePlannerMapChunkRequest, IMessage> {
+        @Override
+        public IMessage onMessage(MessageZonePlannerMapChunkRequest message, MessageContext ctx) {
+            ctx.getServerHandler().playerEntity.mcServer.addScheduledTask(() -> {
+                ZonePlannerMapDataServer.instance.getChunk(ctx.getServerHandler().playerEntity.worldObj, message.chunkX, message.chunkZ, zonePlannerMapChunk -> {
+                    BCMessageHandler.netWrapper.sendTo(new MessageZonePlannerMapChunkResponse(message.chunkX, message.chunkZ, zonePlannerMapChunk), ctx.getServerHandler().playerEntity);
+                });
+            });
```
AlexIIL: If you use the BCMessageHandler to register it (see my earlier comment) then this method is run on the main thread (you get the above code for free).

common/buildcraft/robotics/ZonePlannerMapDataClient.java
```diff
@@ -0,0 +1,28 @@
...
+        Pair[] pendingRequestsArray = new Pair[]{null};
+        pendingRequest = Pair.of(chunkPosPair, zonePlannerMapChunk -> {
+            pendingRequests.remove(pendingRequestsArray[0]);
+            callback.accept(zonePlannerMapChunk);
+        });
+        pendingRequestsArray[0] = pendingRequest;
+        pendingRequests.add(pendingRequest);
+        BCMessageHandler.netWrapper.sendToServer(new MessageZonePlannerMapChunkRequest(chunkX, chunkZ));
+    }
+}
```
AlexIIL: Instead of ````Pair<Integer, Integer>```` you should use ````ChunkPos```` as that's a better class for it.

You *might* want to consider changing this to
````java
private final Map<ChunkPos, Deque<Consumer<ZonePlannerMapChunk>> requestors = new HashMap<>();
````
Where you add requesters to a ````ConcurrentLinkedDeque<>````, if the chunk doesn't exist. (I say concurrent mostly so that we can look into threading the chunk rendering later, although it could be changed if its just an unnecessary addition).

File: common/buildcraft/builders/tile/TileArchitect_Neptune.java [hunk 4]
```diff
@@ -50,34 +52,31 @@
...
     public static final int NET_SCAN = 21;
-    private static final int BLOCKS_PER_TICK = 1;
 
-    protected final IItemHandlerModifiable invBptIn = addInventory("bptIn", 1, EnumAccess.INSERT, EnumPipePart.VALUES);
-    protected final IItemHandlerModifiable invBptOut = addInventory("bptOut", 1, EnumAccess.EXTRACT, EnumPipePart.VALUES);
+    public final IItemHandlerModifiable invBptIn = addInventory("bptIn", 1, EnumAccess.INSERT, EnumPipePart.VALUES);
+    public final IItemHandlerModifiable invBptOut = addInventory("bptOut", 1, EnumAccess.EXTRACT, EnumPipePart.VALUES);
 
     public boolean shouldScanEntities = false;
     /** Details- if true then this will create a blueprint, otherwise it will be a template. */
```
afdw: What is template? Is it like blueprint, but stores only blocks existence, not type?

AlexIIL: A template stores each block as either air, or a full block.

afdw:
https://github.com/BuildCraft/BuildCraftAPI/pull/3

AlexIIL:
I'd rather that you leave the blueprint + template code alone, as I was already planning on making some other changes to its code. (Sorry for not realising this when you asked earlier)

afdw:
What's with zone planner? Editing regions is now working, but they aren't saving on server / in item.

common/buildcraft/builders/tile/TileArchitect_Neptune.java
```diff
@@ -113,26 +111,33 @@ public void onPlacedBy(EntityLivingBase placer, ItemStack stack) {
...
     @Override
     public void update() {
+        deltaProgress.tick();
+
         if (worldObj.isRemote) {
             return;
         }
         if (scanning && isValid) {
-            scanMultipleBlocks();
+            int target = shouldScanDetails ? 30 : 10;
```
AlexIIL: If the architect takes any more than a single tick for a block then its going to take a huge amount of time to scan the entire blueprint, which is something I'd like to avoid.

common/buildcraft/lib/misc/BoxIterator.java
```diff
@@ -92,4 +137,19 @@ private boolean shouldReset(BlockPos current, EnumFacing facing) {
...
+
+    @Override
+    public void writeToByteBuf(ByteBuf buf) {
+        NetworkUtils.writeBlockPos(buf, min);
+        NetworkUtils.writeBlockPos(buf, max);
+        buf.writeBoolean(invert);
+        buf.writeBoolean(repeat);
+        order.writeToByteBuf(buf);
+        NetworkUtils.writeBlockPos(buf, current);
+    }
```
AlexIIL: It would probably be better to calculate the progress on the server and send over the single int rather than sending over the entire box iterator (so can you remove the read/write methods from this class)

common/buildcraft/robotics/ZoneChunk.java
```diff
@@ -4,24 +4,29 @@
...
 import net.minecraft.util.math.BlockPos;
 
-import buildcraft.api.core.ISerializable;
-import buildcraft.core.lib.utils.BitSetUtils;
-import buildcraft.core.lib.utils.NetworkUtils;
+import java.util.BitSet;
+import java.util.Random;
 
-public class ZoneChunk implements ISerializable {
+public class ZoneChunk implements INetworkLoadable_BC8 {
```
AlexIIL: Your misusing ````INetworkLoadable_BC8```` slightly - can you add the generic type so we get the proper type returned?

common/buildcraft/robotics/ZonePlan.java
```diff
@@ -4,28 +4,36 @@
...
 import net.minecraftforge.common.util.Constants;
 
-import buildcraft.api.core.ISerializable;
-import buildcraft.api.core.IZone;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+import java.util.Set;
+
+public class ZonePlan implements IZone, INetworkLoadable_BC8 {
```
AlexIIL: Again, can you add generics to this?

common/buildcraft/robotics/ZonePlannerMapChunk.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class ZonePlannerMapChunk {
+    public Map<BlockPos, Integer> data = new HashMap<>();
+
+    public ZonePlannerMapChunk load(World world, ChunkPos chunkPos) {
+        data.clear();
+        Chunk chunk = world.getChunkFromChunkCoords(chunkPos.chunkXPos, chunkPos.chunkZPos);
+        for(int x = 0; x < 16; x++) {
+            for(int z = 0; z < 16; z++) {
+                BlockPos pos = new BlockPos(x, 0xFF, z);
```
AlexIIL: You can use ````world.getTopSolidOrLiquidBlock(new BlockPos(x, 0, z))```` to find the toppermost block instead of manually traversing.

common/buildcraft/robotics/tile/TileZonePlanner.java
```diff
@@ -0,0 +1,18 @@
+package buildcraft.robotics.tile;
+
+import buildcraft.lib.tile.TileBCInventory_Neptune;
+import buildcraft.lib.tile.item.ItemHandlerManager;
+import buildcraft.lib.tile.item.ItemHandlerSimple;
+import buildcraft.robotics.ZonePlan;
+
+public class TileZonePlanner extends TileBCInventory_Neptune {
+    public final ItemHandlerSimple invPaintbrushes;
```
AlexIIL: Why do paintbrushes need their own inventory? Can you just manually render them all out in the gui?

afdw:
I reverted quarry commits.

AlexIIL:
Thanks :)

PR #3419, opened 2016-08-25
Quarry

afdw:


common/buildcraft/lib/misc/CullTESR.java
```diff
@@ -0,0 +1,62 @@
...
+
+        VertexBuffer.begin(GL11.GL_QUADS, DefaultVertexFormats.BLOCK);
+
+        renderTileEntityFast(tile, x, y, z, partialTicks, destroyStage, VertexBuffer);
+        VertexBuffer.setTranslation(0, 0, 0);
+
+        tessellator.draw();
+
+        RenderHelper.enableStandardItemLighting();
+    }
```
AlexIIL: Unfortunately this won't quite work - FastTESR is fast because forge has a global VertexBuffer (in ````TileEntityRenderBuffer````). This method is only called by non-fast TESR stuffs.

afdw: It's non-fast TESR.

AlexIIL: ...oh. In which case it can probably just be a normal TESR rather than implement the fast method.

afdw: What is difference? `hasFastRenderer` in `TileTank` is false, so it's just normal TESR.

common/buildcraft/builders/tile/TileQuarry.java
```diff
@@ -291,6 +296,20 @@ public boolean hasCapability(Capability<?> capability, EnumFacing facing) {
         return super.getCapability(capability, facing);
     }
 
+
+
+    @Override
+    @SideOnly(Side.CLIENT)
+    public AxisAlignedBB getRenderBoundingBox() {
+        return box == null ? super.getRenderBoundingBox() : BoundingBoxUtil.makeFrom(getPos(), box);
```
AlexIIL: BoundingBoxUtil checks to see if box is null, and if it is it just uses "pos". So you don't need the null check, and can replace this with
````java
public AxisAlignedBB getRenderBoundingBox() {
    return BoundingBoxUtil.makeFrom(getPos(), box);
}
````
and it won't ever throw an NPE.

AEnterprise:
can i sugest that eighter @afdw squaches his commit before merging or merging with a squach using github? this is already up to 36 commit and after the merge it would be kinda hard to trace specific parts as they will be in between the already pushed commits

afdw:
@AEnterprise I can't understand. Why someone need to trace parts? What are parts? What I do wrong?

AEnterprise:
@afdw you're not doing anything wrong, while you are working it's fine to make individual little commits but to find the changes later a single commit is easier, but github can take care of that

PR #3463, opened 2016-12-04
Fix guide always logging

afdw:


afdw:
@AlexIIL Can you get on IRC please?

PR #3466, opened 2016-12-05
Misc fixes

afdw:


AEnterprise:
this could probably use a better title then "8.0.x" as that realy doesn't say anything about what this PR is about

AlexIIL:
Looks good, but can you check to see why the gui is darker than it should be? (specifically the world, I think it should be at full brightness rather than lower)
screenshot from 2016-12-09 20 49 38

afdw:
@AlexIIL This is probably because you have other blocks in world that break OpenGL settings?

AlexIIL:
That makes sense

PR #3467, opened 2016-12-05
Update bottom of zone planner to match retexture

LuigiHutch:
Brings bottom of zone planner texture more inline with current retexture. Texture taken from the electronic library


AlexIIL:
This makes a lot of sense -- thanks!

Although I'm going to need to check how the BC assets are licensed as they were special cased in the original relicense.

Issue #3479, opened 2017-01-12
[8.0 Alpha Build 2] Shift-Clicking in a BuildCraft Block crashes game


NinjaKap:
Hello!
I found the new Alpha Build 2 for BC 8.0.0, and decided to give it a shot.

The Engine Texture is glitched and names messed up, but still functional. I placed one down, and shift-clicked coal blocks in; it crashed my game.

I did the same in a combustion engine (to move around inv), and crashed game.

I tried again in the Quarry Block, and it, again, crashed.

Not sure if you're "looking for" bugs yet, just thought I'd put this here (in-case you didn't know!)

Thanks for all you've done, and can't wait to mess around with the final release!

afdw:
> I tried again in the Quarry Block, and it, again, crashed.
When? What you was doing?

NinjaKap:
I opened the Quarry (as it now as a GUI in 8.0), and wondered if the new items are a filter. I couldn't put anything in, and tried shift-clicking, when it crashed.

afdw:
You should put frames in it

NinjaKap:
Ah, okay.

forReason:
as seen in the reference above, the issue not only affects engines but every shift-click interaction. #3468 were two issues so i referenced the shift click here to get a clean, separate issue

AlexIIL:
Should be fixed by the commits mentioned above.

PR #3490, opened 2017-01-27
Fixed Shift-Click crash (#3479)

david-simon:
I think I fixed the crashes caused by Shift-Clinking in any BC Container.

The crash is caused by the slotClick() function ofnet.minecraft.inventory.Container
which calls transferStackInSlot(). ContainerBC_Neptune didn't override this method, the default implementation was called which basicly does nothing. After this, slotClick() checks whether the transfer was successful and retrys if not. Since the transfer didn't happen it falls into an infinite loop and crashes.

ContainerEngineStone_BC8 still crashed because it only checked if an item is valid for a slot when it was already being inserted ( see ItemHandlerSimple, setStackInSlot() ) so I added isValidItem() to it's SlotBase implementation.

### DIFF
--- a/common/buildcraft/energy/container/ContainerEngineStone_BC8.java
+++ b/common/buildcraft/energy/container/ContainerEngineStone_BC8.java
@@ -6,6 +6,8 @@ import net.minecraft.inventory.IContainerListener;
 import buildcraft.energy.tile.TileEngineStone_BC8;
 import buildcraft.lib.gui.ContainerBCTile;
 import buildcraft.lib.gui.slot.SlotBase;
+import net.minecraft.item.ItemStack;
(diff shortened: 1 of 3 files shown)


david-simon:
Oops, didn't mean to push the AutoWorkbench stuff

AEnterprise:
travis failure is travis getting a timeout when trying to setup the workspace so can be ignored, original commit does build

AlexIIL:
Until we've fully relicensed to MPLv2 can you agree to the [relicense](https://github.com/BuildCraft/BuildCraft/issues/3004)? Its kind of in limbo between licenses atm, so I don't think its auto-covered.

david-simon:
I agreed to the relicense

AlexIIL:
ok, thanks

Issue #3468, opened 2016-12-17
[8.0.x] Auto workbench Duplication


kruthers:
When crafting in a auto workbench any recipy will only require 1 item to craft. E.g. if crafing a diamond block in the workbench once you make the ghost recipy you only need to put in 1 diamond to start the carfting process alowing you to get 8 extra diamonds for free.

This also works with complex recipies (ones that require more then 1 item). E.g. if crafting a auto workbench in a auto workbench all you would need to put in is 1 gear and 1 crafting table instead of 2 gears and 1 crafting table.

kruthers:
The game also crashes when you shift click items out of them into you inventory

AEnterprise:
8.0.x is not released yet officially so this might be fixed or not, please provide the commit used to build your version so we can see if this is fixed or not

AlexIIL:
Yeah I broke autoworkbenches sometime ago and never got around to fixing them

forReason:
The shiftclick also crashes the game when trying to put any fuel into the steamengine this probably refers to the second issue 

"The game also crashes when you shift click items out of them into you inventory"

AlexIIL:
shiftclicking crashes every BC inventory atm.

forReason:
Shiftclicking bug split into seperate issue:  #3479

PR #3491, opened 2017-01-29
Fixed AutoWorkbench item duplication (#3468).

david-simon:
IRecipe.matches ( called from hasMaterialsForRecipe() ) uses CraftSlotItemBound.get(). Since it only simulated extracting items from the materials inventory, it returned an ItemStack even if it has been already used. My solution was to copy the materials inventory before applying the bindings and actually extracting from it.

I found some issues with how itemStackcache caused by onSlotChange() behaviour and ItemStackKey comparisons so I added a method to get the key of a given ItemStack which only contains one item and that seemed to solve those issues.

I'm honestly not too happy with how the rebind() method turned out so please do tell me if there's a cleaner solution.

### DIFF
--- a/common/buildcraft/factory/tile/TileAutoWorkbenchBase.java
+++ b/common/buildcraft/factory/tile/TileAutoWorkbenchBase.java
@@ -47,6 +47,7 @@ public abstract class TileAutoWorkbenchBase extends TileBC_Neptune implements IT
     public final ItemHandlerSimple invResult;
     public final ItemHandlerSimple invOverflow;
     protected final Map<ItemStackKey, TIntHashSet> itemStackCache;
+    private ItemHandlerSimple invMaterialsCopy = null;
 
     public IRecipe currentRecipe;
     private int progress = -1;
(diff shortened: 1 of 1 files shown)


AEnterprise:
pinging @Flow86 for the weird jenkins failure

david-simon:
Jenkins might have failed because my repo was behind, the error references a BuildCraftAPI commit. Could have been caused by 4ab8b2b53df532e8c654ad179d8d0ee1f4741e7b or 65ae2f7983361833cf3730b0fa8ed359afbdb218 but it should be good now.

AlexIIL:
It might be cleaner to pull all items into a copied workbench before crafting and do away with the bindings. I'm not really sure though - it really depends on which is simpler.

AEnterprise:
@david-simon it failed again, similar error

@AlexIIL i'm no fan of copying and removing bindings, it's easy to go wrong and have item dupes

david-simon:
@AEnterprise the error is referencing the api submodule again. There were a bunch of commits pushed since this PR so it is behind again,  but I don't think it's anything serious

@AlexIIL I'll have a look

AEnterprise:
it has merge conflics that need to be resolved before this can be tested and merged

Prototik:
Can one of the admins verify this patch?

afdw:
@AlexIIL will do after my work on builders.

Prototik:
Can one of the admins verify this patch?

AEnterprise:
@david-simon any update on the conlicts?

david-simon:
I will look into it tomorrow

2017. máj. 19. 10:59 ezt írta ("AEnterprise" <notifications@github.com>):

david-simon:
@AEnterprise I think it's finally resolved

AEnterprise:
it is a valid commit ref but it's not the latest one, need to fix the travis script for that

AEnterprise:
i stand corrected: not a travis issue, the commit was valid as the last commit in the latest PR, that got merged with a squash, changing the commit hash

AEnterprise:
closing this PR as i rewrote the way the auto workbench works making this no longer needed

but thanks for tracing this earlier and making this PR, it showed me where it went wrong so i could avoid the same mistake in this new version

PR #3587, opened 2017-05-17
7.2.x Even item distribution on split for Quartz pipe

dsmiller95:
This change will modify the behavior of the quartz pipe, so that when items come through it in a split they are deterministically routed to the output pipes to guarantee an even distribution of items between all of the output pipes.

It abstracts the routing logic out of the PipeLogicIron and extends it to a new PipeLogicQuartz class to allow it to route the items that come through. In addition, a new method is added to the Pipe class to allow a pipe to modify its state before each item is routed through, which is necessary for this type of behavior without making much larger changes.

I thought this would be useful for larger factories, especially in loop-based configurations where maintaining a perfectly even split would be useful; while not changing the base functionality of the quartz pipe.


asiekierka:
I'm afraid 7.2.x is not maintained in terms of feature additions, or even bugfixes. In addition, this kind of feature is breaking to all existing builds, so we cannot just add it in a minor release. For that reason, if anything, it would have to be targetted to 8.0.x.

I am aware that there's a fair amount of effort in this PR, but I don't think we can make it work.

Also, did you really need to add new methods? The diamond pipe already performed weighted, round-robin/even item distribution on split for all matching filters. You could probably make a pipe addon which adds an evenly distributing pipe based on the techniques contained therein.

dsmiller95:
Wow, I feel like an idiot. I had no idea the diamond pipe already implemented this kind of behavior. thanks for the suggestion and feedback!

asiekierka:
@dsmiller95 Don't feel bad; BuildCraft is not known for good documentation, and we are aware of this - the new team is working on an in-game guidebook for that very reason!

Issue #3604, opened 2017-05-23
Auto workbench not present in game


peterwy22:
BC version 7.99.3 Dev build forge version 13.20.0.2296, Can't find auto workbench in item list as well as it's recipe

AlexIIL:
They aren't yet enabled in the alpha because of #3468.

(I'll keep this open until that one is fixed, as there might be other issues before that one is done. Also because then we have a single issue for autoworkbenches)

PR #3620, opened 2017-05-28
rewrote autocrafting

AEnterprise:


AEnterprise:
should fix folowing issues: #3604 , #3468

buildcraft_resources/assets/buildcraft/lang/en_US.lang
```diff
@@ -478,7 +478,7 @@ itemGroup.buildcraft.plugs=BuildCraft Pluggables
 
 tile.architectBlock.name=Architect Table
 tile.assemblyTableBlock.name=Assembly Table
-tile.assemblyWorkbenchBlock.name=Advanced Crafting Table
+tile.advancedCraftingTableBlock.name=Advanced Crafting Table
```
AlexIIL: Don't change lang keys unless the real name changed.

common/buildcraft/lib/tile/item/ItemHandlerSimple.java
```diff
@@ -235,4 +237,21 @@ private void setStackInternal(int slot, @Nonnull ItemStack stack) {
...
+                return true;
+            if (StackUtil.canMerge(stored, stack)){
+                amount -= stored.getMaxStackSize() - stored.getCount();
+                if (amount <= 0) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
```
AlexIIL: This can be "insert(stack, true, true)" (allAtOnce is true, simulate is true)

buildcraft_resources/assets/buildcraft/lang/en_US.lang
AEnterprise: it did, i didn't change anything and it wasn't localized

common/buildcraft/lib/tile/item/ItemHandlerSimple.java
AEnterprise: alright changed

buildcraft_resources/assets/buildcraft/lang/en_US.lang
AlexIIL: You misunderstand: we shouldn't change lang keys from 1.7.10 unless the thing that they represent has changed its name.

If the code is pointing to the wrong lang key then it should be fixed on the code side. (This saves us from having to change every lang file)

AlexIIL:
According to @LuigiHutch it already works but I'll just test myself to check :)

AlexIIL:
Thanks!

Issue #3601, opened 2017-05-22
bring back oil spillage round oil springs


AEnterprise:
oil wells used to have bits of oil spilled arround the springs, making them feel alot more natural

allong with the fact that you get pushed down and slowed down when swimming in oil this would also make setting up a pump a bit more interesting without flight

![image](https://cloud.githubusercontent.com/assets/5206666/26323936/02686bda-3f31-11e7-95ac-6b9289c01df3.png)

AEnterprise:
this got lost in the 1.6->1.7 port so we can probably reuse code from then (depending on licencing, would need checking)

PR #3626, opened 2017-05-29
fix oil tendrill generation (#3601)

AEnterprise:
base problem was a check that got inverted, also some extra checks for if a block is replacable, tendrills spill into nearby chunks that have not generated yet so the first block(s) eighter return air or flowing water while the generator realises the chunk has not been generated and interups the oil spilage to generate the new chunk


AEnterprise:
not ready to be merged, apearently BC wordgen needs a rewrite to prevent runaway chunk generation

AlexIIL:
We'll fix runaway chunk gen seperatly

PR #3654, opened 2017-06-05
FakeServer, threading and some builder fixes

AEnterprise:
Reworked the fakeworld into a client version (for rendering) and a server version (with attached dimension that could later be used for blueprint editing though that would require fixing up some things, currently you can't place any blocks in the dimension as a player)

functionality remained the same although the computing of requirements on the server needs to be split up over a period of time (with a queue). Currently it locks the server thread until it finishes or the watchdog kills it (depending on the size)

Tested with integrated server running and a dedicated one


common/buildcraft/builders/snapshot/Blueprint.java
```diff
@@ -97,9 +98,6 @@ public void deserializeNBT(NBTTagCompound nbt) throws InvalidInputDataException
         NBTTagList list = nbt.hasKey("data", Constants.NBT.TAG_LIST) ? nbt.getTagList("data", Constants.NBT.TAG_INT)
             : null;
         int[] serializedData = nbt.hasKey("data", Constants.NBT.TAG_INT_ARRAY) ? nbt.getIntArray("data") : new int[0];
-        if (serializedData == null && list == null) {
-            throw new InvalidInputDataException("Can't read a blueprint with no data!");
-        }
```
AlexIIL: ...Don't remove this.

We catch and handle ````IOException```` later down the line, we don't handle NullPointerExceptions however.

AEnterprise: it simply can't be null, if you look at the 2 this is data that already checked in the 2 lines above with nbt.hasKey and a ternairy if

AEnterprise:
i think this is ready now

common/buildcraft/lib/misc/WorkerThreadUtil.java
```diff
@@ -66,29 +66,29 @@
...
     public static void executeWorkTask(Runnable task) {
-        executeWorkTask(new CallableDelegate(task));
+        executeWorkTask(new CallableDelegate(task), true);
     }
 
     /** Executes a task. If this is in debug mode then If this takes longer than 30ms the it assumes that something has
      * gone wrong, and will notify the log that a task took too long. If it goes on for longer than 10s then it will
      * make a big error in the log. */
-    public static <T> Future<T> executeWorkTask(Callable<T> task) {
+    public static <T> Future<T> executeWorkTask(Callable<T> task, boolean shouldMonitor) {
```
AlexIIL: Leave the monitoring in please. - it should never be optional.

common/buildcraft/lib/dimension/BlueprintCalculator.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        fakeServer.clear(location);
+        BlueprintLocationManager.releaseLocation(location);
+        return new BuildingInfoData(Pair.of(requiredItems, requiredFluids), Pair.of(requiredItemsForEntities, requiredFluidsForEntities));
+    }
+
+    public static class BuildingInfoData {
+        public final Pair<List<ItemStack>[][][], List<FluidStack>[][][]> blockRequirements;
+        public final Pair<List<List<ItemStack>>, List<List<FluidStack>>> entityRequirements;
```
AlexIIL: Use separate variables for this, don't use ````Pair````.

common/buildcraft/builders/snapshot/Blueprint.java
AlexIIL: Can you replace ````serializedData == null```` with ````serializedData.length == 0````?

common/buildcraft/builders/snapshot/Blueprint.java
```diff
@@ -137,19 +139,32 @@ public EnumSnapshotType getType() {
...
+            future = WorkerThreadUtil.executeWorkTask(new BlueprintCalculator(Blueprint.this), false);
+        }
+
+        private void processData() {
+            BlueprintCalculator.BuildingInfoData infoData = null;
+            try {
+                infoData = future.get();
+            } catch (InterruptedException | ExecutionException e) {
+                //Future reported to be done and we are not canceling so this should never heapen
+                e.printStackTrace();
```
AlexIIL: If an error should never happen then we probably need to know about it: Please throw a ````RuntimeException```` or an ````Error```` instead of discarding the ````ExecutionException````. I'm not sure what the best behaviour is for handling  ````InterruptedException```` though.

common/buildcraft/builders/snapshot/Blueprint.java
```diff
@@ -137,19 +139,32 @@ public EnumSnapshotType getType() {
...
+            }
+            Pair<List<ItemStack>[][][], List<FluidStack>[][][]> required = infoData.blockRequirements;
+            Pair<List<List<ItemStack>>, List<List<FluidStack>>> requiredEntities = infoData.entityRequirements;
```
AlexIIL: Don't use ````Pair````

common/buildcraft/builders/tile/TileBuilder.java
```diff
@@ -246,6 +247,7 @@ public void writePayload(int id, PacketBufferBC buffer, Side side) {
         }
     }
 
+    @SideOnly(Side.CLIENT)
```
AlexIIL: This shouldn't be marked as ````@SideOnly(Side.CLIENT)````

common/buildcraft/lib/dimension/BlankWorldProvider.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public boolean hasNoSky() {
+        return super.hasNoSky();
+    }
+
+    @Override
+    public String getSaveFolder() {
+        return "The dimension of blueprints";
```
AlexIIL: ....Why that string?

If it shouldn't be saved then please make it obviously wrong: something like ````DIM_BC_ERROR````

common/buildcraft/lib/dimension/FakeWorldServer.java
```diff
@@ -0,0 +1,302 @@
...
+import buildcraft.api.schematics.SchematicBlockContext;
+
+import buildcraft.lib.BCLib;
+import buildcraft.lib.misc.data.Box;
+
+public class FakeWorldServer extends WorldServerMulti implements IFakeWorld {
+    private List<ItemStack> drops = new ArrayList<>();
+    public static FakeWorldServer INSTANCE;
+    private FakeChunkProvider provider;
+    private List<Box> locks = new ArrayList<>();
```
AlexIIL: If you access this list from (potentially) multiple threads then you're going to need to make it threadsafe.

common/buildcraft/lib/misc/WorkerThreadUtil.java
AEnterprise: alright just thought this would be the least invasive option to not have my task killed

common/buildcraft/lib/dimension/BlueprintCalculator.java
AEnterprise: alright, the original code used pairs, i'll change that to not need it then

common/buildcraft/builders/snapshot/Blueprint.java
AEnterprise: interrupted i would ignore in any case as that is what gets thrown if it gets interuped before finishing but that cannot heapen as long as the monitor doesn't kick in

common/buildcraft/builders/tile/TileBuilder.java
AEnterprise: it uses a client side only method still, tankManager.readData

i'll remove the annotation but if this gets called on a server it will crash on that then

common/buildcraft/lib/dimension/BlankWorldProvider.java
AEnterprise: i wouldn't call it error though, due to how MC works this folder will always be generated with a few chunks. I cannot realy avoid that

common/buildcraft/builders/tile/TileBuilder.java
AlexIIL: The argument "side" is guaranteed to never be CLIENT on the server. So its safe.

Issue #3613, opened 2017-05-25
7.99.4: Gates do note out put a redstone signal


Jigokusaru:
![screenshot1](https://cloud.githubusercontent.com/assets/23660999/26468381/34a7ad1a-4164-11e7-911d-201cbd3b9e22.png)
![screenshot2](https://cloud.githubusercontent.com/assets/23660999/26468596/10d550d0-4165-11e7-809a-eeaee008afc9.png)

The gate does not out put a signal even though conditions are met and redstone is selected as the action.

Elec332:
This is still an issue on 7.99.4

AEnterprise:
yeah nobody looked at this yet, i might once i'm done with tinkering with the builder

AlexIIL:
Fixed in [7.99.5](https://www.mod-buildcraft.com/buildcraft-7995-alpha.html)

PR #3694, opened 2017-06-28
make gates emit redstone again

AEnterprise:
implements #3613 again

### DIFF
--- a/common/buildcraft/transport/block/BlockPipeHolder.java
+++ b/common/buildcraft/transport/block/BlockPipeHolder.java
@@ -10,7 +10,6 @@ import java.lang.ref.WeakReference;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
-
 import javax.annotation.Nullable;
 
 import net.minecraft.block.Block;
@@ -66,6 +65,7 @@ import buildcraft.lib.misc.VecUtil;
 import buildcraft.lib.prop.UnlistedNonNullProperty;
 
 import buildcraft.transport.BCTransportItems;
+import buildcraft.transport.BCTransportPlugs;
 import buildcraft.transport.item.ItemWire;
 import buildcraft.transport.pipe.Pipe;
 import buildcraft.transport.tile.TilePipeHolder;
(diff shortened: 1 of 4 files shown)


common/buildcraft/transport/tile/TilePipeHolder.java
```diff
@@ -72,6 +72,9 @@
     public static final int NET_UPDATE_PLUG_EAST = getReceiverId(PipeMessageReceiver.PLUGGABLE_EAST);
     public static final int NET_UPDATE_WIRES = getReceiverId(PipeMessageReceiver.WIRES);
 
+    private int[] redstoneValues = new int[6];
+    private int[] oldredstoneValues = new int[]{ -1, -1, -1, -1, -1, -1};
```
AlexIIL: Make these final, and copy over the values in readFromNbt (in case of bad nbt)

common/buildcraft/transport/block/BlockPipeHolder.java
```diff
@@ -725,4 +725,43 @@ public IBlockState getExtendedState(IBlockState state, IBlockAccess world, Block
...
     }
+
+    @Override
+    public boolean canConnectRedstone(IBlockState state, IBlockAccess world, BlockPos pos, @Nullable EnumFacing side) {
+        if (side == null)
+            return false;
+        TilePipeHolder tile = getPipe(world, pos, false);
+        if (tile != null) {
+            PipePluggable pluggable = tile.getPluggable(side.getOpposite());
+            return pluggable != null && pluggable.definition == BCTransportPlugs.gate;
```
AlexIIL: Is there some way to make this not hardcoded? I'd like to support other pluggables or even other types of PipeBehaviour that can connect to redstone.

common/buildcraft/transport/tile/TilePipeHolder.java
AEnterprise: in case of bad nbt it won't matter realy if i assign or copy over, it's pretty much impossible to protect against nbt being corrupted or tampered with

PR #3698, opened 2017-07-01
misc fixes

AEnterprise:
combustion engine outputs now correctly reset when switching fuels
quarry isn't constantly rebuilding the list of frame locations
quarry now returns if a block gets placed in a prev dug location
quarry chunkloads it's area again
quarry makes the anvil sound again
quarry no longer steals player items using item capabilities
quarry picks up all dropped items again

### DIFF
--- a/common/buildcraft/builders/tile/TileQuarry.java
+++ b/common/buildcraft/builders/tile/TileQuarry.java
@@ -13,9 +13,10 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.function.Function;
 import java.util.function.Predicate;
-
 import javax.annotation.Nonnull;
 
 import com.google.common.collect.ImmutableList;
@@ -87,6 +88,9 @@ public class TileQuarry extends TileBC_Neptune implements ITickable, IDebuggable
      * Recent power input, in MJ (not micro)
      */
     private final AverageInt recentPowerAverage = new AverageInt(200);
+
+    private List<BlockPos> frameLocations;
+    private final Set<BlockPos> breakPoses = new TreeSet<>(Comparator.comparingDouble(o -> o.distanceSq(pos)));
 
     public TileQuarry() {
         caps.addProvider(new MjCapabilityHelper(new MjBatteryReciver(battery) {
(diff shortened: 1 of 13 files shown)


common/buildcraft/lib/block/BlockBCTile_Neptune.java
```diff
@@ -42,6 +47,9 @@ public void breakBlock(World world, BlockPos pos, IBlockState state) {
             TileBC_Neptune tileBC = (TileBC_Neptune) tile;
             tileBC.onRemove();
         }
+        if (tile instanceof IChunkLoadingTile) {
+            ChunkLoaderManager.releaseChunksFor((IChunkLoadingTile) tile);
+        }
```
AlexIIL: This should be in ````TileQuarry.invalidate````.

common/buildcraft/builders/tile/TileQuarry.java
```diff
@@ -635,6 +673,20 @@ public final boolean addPower(long microJoules) {
                 return onReceivePower();
             }
         }
+
+        public List<ItemStack> pickupItems(BlockPos pos) {
+            List<ItemStack> stacks = new ArrayList<>();
+            for (Entity entity : world.getEntitiesWithinAABB(Entity.class, new AxisAlignedBB(pos).expandXyz(1))) {
+                if (entity instanceof EntityPlayer)
+                    continue;
```
AlexIIL:
````java
for (EntityItem entity : world.getEntitiesWithinAABB(EntityItem.class, new AxisAlignedBB(pos).expandXyz(1))) {
  // Rest of code
}
````
As we don't want to be stealing items from anything other than a dropped item.

This could also just call ````InventoryUtil.addToBestAcceptor```` manually, as that's what all callers do (no need to add to a list).

File: common/buildcraft/lib/chunkload/IChunkLoadingTile.java [hunk 3]
```diff
@@ -20,16 +19,20 @@
 import buildcraft.lib.BCLibConfig;
 import buildcraft.lib.BCLibConfig.ChunkLoaderLevel;
 
-/** This should be implemented by {@link TileEntity}'s that wish to be chunkloaded by buildcraft lib. Note that tiles
- * should add themselves to the chunkloading list in {@link ChunkLoaderManager#loadChunksForTile(IChunkLoadingTile)} */
+/**
+ * This should be implemented by {@link TileEntity}'s that wish to be chunkloaded by buildcraft lib.
+ */
```
AEnterprise: guess intellij reformated that when i opened the file for some reason

common/buildcraft/lib/block/BlockBCTile_Neptune.java
AEnterprise: right will move

PR #3702, opened 2017-07-05
quarry fixes

AEnterprise:
remake of #3698

brings back quarry chunkloading and a few minor optimalizations


common/buildcraft/builders/tile/TileQuarry.java
```diff
@@ -325,17 +331,46 @@ public void invalidate() {
...
+
+    private void loadFrameCaches() {
+        IBlockState state = world.getBlockState(pos);
+        if (state.getBlock() != BCBuildersBlocks.quarry) {
+            frameLocations = Collections.emptyList();
         }
         toCheck.clear();
-        toCheck.addAll(frameBoxPoses);
+        if (frameBox.isInitialized()) {
+            toCheck.addAll(frameBox.getBlocksInArea());
```
AEnterprise: but we don't want the edges, we want to check not only the frame locations but the entire box encompased by the frames has to be empty for the drill to have room to work

PR #3706, opened 2017-07-07
Fix non-square recipes exploding GuideCraftingFactory

Chocohead:
Fixes crashes like https://ghostbin.com/paste/ycfse

### DIFF
--- a/common/buildcraft/lib/client/guide/parts/recipe/GuideCraftingFactory.java
+++ b/common/buildcraft/lib/client/guide/parts/recipe/GuideCraftingFactory.java
@@ -128,15 +128,16 @@ public class GuideCraftingFactory implements GuidePartFactory {
             ChangingItemStack[][] inputGrid = { input };
             if (recipe instanceof IRecipeViewable.IViewableGrid) {
                 int width = ((IRecipeViewable.IViewableGrid) recipe).getRecipeWidth();
-                inputGrid = new ChangingItemStack[inputGrid.length / width][width];
+                int height = ((IRecipeViewable.IViewableGrid) recipe).getRecipeHeight();
+                inputGrid = new ChangingItemStack[width][height];
                 int x = 0;
                 int y = 0;
                 for (ChangingItemStack anInput : input) {
                     inputGrid[x][y] = anInput;
-                    x++;
-                    if (x > width) {
-                        x = 0;
-                        y++;
+                    y++;
+                    if (y >= height) {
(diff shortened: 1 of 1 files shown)


common/buildcraft/lib/client/guide/parts/recipe/GuideCraftingFactory.java
```diff
@@ -128,15 +128,16 @@ public static GuidePartFactory getFactory(IRecipe recipe) {
...
                 for (ChangingItemStack anInput : input) {
                     inputGrid[x][y] = anInput;
-                    x++;
-                    if (x > width) {
-                        x = 0;
-                        y++;
+                    y++;
+                    if (y >= height) {
+                        y = 0;
+                        x++;
```
AlexIIL: I thought recipes were given row 1, row 2, etc, rather than column 1, column 2, etc. (I'll test this when I'm back at my dev environment though)

Chocohead: That's what I thought too, but trying it out it seems the guide does them by column

AlexIIL: Sounds like another bug somewhere else in the guide code... what recipes do this wrong? (Just so I can start debugging somewhere)

Chocohead: `GuideCrafting#renderIntoArea` looks to be the cause of it working by column instead of row, it consistently iterates each `y` for every `x` rather than the other way around.

As for the recipe crash, there's nothing natively in BuildCraft that causes it at the moment because the only thing that implements `IViewableGrid` is the facade swapping recipes, which are 1x1. That's probably why this hasn't come up before. I was making a version of IC2's `AdvRecipe` that also implemented `IViewableGrid`, but something like [this](https://gist.github.com/Chocohead/911420cf88df2580ac7639ff99e46903) should show what happens.

Chocohead:
All good now with 7.99.6, thanks

AlexIIL:
...Sorry for not replying, I had a draft and forgot to press send :(

AlexIIL:
In essence, thanks for reporting this ( :D ), but its so small that its easier to fix it myself than request that you change your PR

Chocohead:
That's fine, it's more important that it works than how it gets there ;)

Issue #3721, opened 2017-07-11
[1.11.2] Crash when looking at Auto Workbench in F3 Debug Mode


stdio2016:
Steps to Reproduce:
1. Create a single player map
2. Enter F3 Debug Mode
3. Build an Auto Workbench
4. Look at it
5. `Shutting down internal server`, then crash

---- Minecraft Crash Report ----
// Why did you do that?

Time: 7/11/17 8:41 AM
Description: Unexpected error

java.lang.NullPointerException: Unexpected error
	at buildcraft.factory.tile.TileAutoWorkbenchBase.getDebugInfo(TileAutoWorkbenchBase.java:82)
	at buildcraft.core.client.RenderTickListener.renderOverlay(RenderTickListener.java:125)
	at net.minecraftforge.fml.common.eventhandler.ASMEventHandler_35_RenderTickListener_renderOverlay_Text.invoke(.dynamic)
	at net.minecraftforge.fml.common.eventhandler.ASMEventHandler.invoke(ASMEventHandler.java:90)
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:185)
	at net.minecraftforge.client.GuiIngameForge.renderHUDText(GuiIngameForge.java:695)
	at net.minecraftforge.client.GuiIngameForge.renderGameOverlay(GuiIngameForge.java:171)
	at net.minecraft.client.renderer.EntityRenderer.updateCameraAndRender(EntityRenderer.java:1089)
	at net.minecraft.client.Minecraft.runGameLoop(Minecraft.java:1077)
	at net.minecraft.client.Minecraft.run(Minecraft.java:372)
	at net.minecraft.client.main.Main.main(SourceFile:124)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at net.minecraft.launchwrapper.Launch.launch(Launch.java:135)
	at net.minecraft.launchwrapper.Launch.main(Launch.java:28)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #3723, opened 2017-07-11
Update TileAutoWorkbenchBase.java

AlexIIL:


AlexIIL:
Fixes #3721

AlexIIL:
(Done as a PR so that travis can check to see if I've messed something up)

AEnterprise:
fine but you are suposed to be enjoying your time away, i fixed this on 1.12 already and was gona backport any fixes i do in the next weeks all at once

so if you realy want to fix things without dev environment go ahead but make sure you don't do to much or worry about the bugs in your off time

AlexIIL:
You *almost* fixed it - it could have still thrown an NPE when running the integrated server.

Don't worry about my holiday - it has plenty of downtime when I've got nothing better to do.

AEnterprise:
my fix was basicly a null check so wouldn't have crashed but this fix is fine as well

AlexIIL:
The issue with your fix is that you checked to see if it was null, then used the variable if it wasn't null. This would be fine in a single threaded environment, but in multithreaded environments (like when the client thread is reading data from the server thread) the server thread might have changed the variable in between the check and the access. (It seems unlikely, but it can happen. It's more likely to happen with other blocks that change between null and a value each tick, especially as it's being called every frame)

AlexIIL:
(This is also why I wanted to fix this myself while the problem was fresh in my mind - it's unlikely that's I would have noticed the issue with your fix at first, as it looks fine by default)

AEnterprise:
oh right, well thanks for spotting then :P

PR #3753, opened 2017-07-22
Backport fixes

AEnterprise:
backports some of the fixes i did while working on 1.12

### DIFF
--- a/common/buildcraft/energy/generation/GenLayerAddOilDesert.java
+++ b/common/buildcraft/energy/generation/GenLayerAddOilDesert.java
@@ -14,10 +14,6 @@ import net.minecraft.world.gen.layer.GenLayer;
 
 import net.minecraftforge.common.BiomeDictionary;
 
-import buildcraft.api.core.BCLog;
-
-import buildcraft.lib.BCLib;
-
 public class GenLayerAddOilDesert extends GenLayerBiomeReplacer {
     private static final double NOISE_FIELD_SCALE = 0.001;
     private static final double NOISE_FIELD_THRESHOLD = 0.7;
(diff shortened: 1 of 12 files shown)


AEnterprise:
added a fix for redstone engines outputting to things they should not output (like wooden kinesis pipes)

common/buildcraft/factory/tile/TileMiner.java
```diff
@@ -92,6 +92,11 @@ public void update() {
     }
 
     @Override
+    public void onLoad() {
+        offset = world.rand.nextInt(10);
```
AlexIIL: Add 
````java
super.onLoad();
````
Just so that we don't have to worry about future additions to the parent's onLoad function not being called by this.

File: common/buildcraft/lib/engine/TileEngineBase_BC8.java [hunk 2]
```diff
@@ -376,7 +376,7 @@ private void sendPower() {
             return;
         }
         IMjReceiver receiver = getReceiverToPower(tile, currentDirection);
-        if (receiver != null) {
+        if (receiver != null && mjConnector.canConnect(receiver) && receiver.canConnect(mjConnector)) {
```
AlexIIL: This check is partially implemented in getReceiverToPower -- the addition should go there.

common/buildcraft/energy/generation/GenLayerAddOilDesert.java
```diff
@@ -34,7 +36,8 @@ public GenLayerAddOilDesert(long worldSeed, long seed, GenLayer parent) {
     protected boolean canReplaceBiome(int biomeId) {
         Biome biome = Biome.getBiomeForId(biomeId);
         if (biome == null) {
-            BCLog.logger.warn("Couldn't find a biome with an ID of " + biomeId);
+            if (BCLib.DEV) //don't spam outside of dev, it's way to noisy
+                BCLog.logger.warn("Couldn't find a biome with an ID of " + biomeId);
```
AlexIIL: Just removing the log would be fine actually, we don't have any use for it in dev

Issue #3513, opened 2017-03-29
Suggestion


lllMohandarlll:
Hello, I'm actually working to create a pack centred about realism/educational aspect, tweaking around when necessary through minetweaker and other stuffs.
While I find cool the laser animation in buildcraft, normal industrial lasers operate at light frequencies not visible for the human eye.
My suggestion was: make it possible to have an item to craft (likely particular glasses) in order to be able to see the render animation generated by bc lasers.
This, naturally not necessarily as main feature, but maybe as something possible to enable as an option in the config file.

AEnterprise:
the problem with changing the way the lasers work is that they have always worked in the current visible way, if suddenly BC lasers stop showing laser beams then that would cause alot of confusion

if this would be done it would need a config option to be dissabled, particle effects on the table itself is tricky as when people have graphics settings turned down they might not be visible

LuigiHutch:
Make it disabled by default so it is only available with a config changed. With the config enabled, they are visible if they are wearing the ~~robot~~ goggles (not sure if they are still planned) and add this info to the guide and/or tooltip. Otherwise with this config disabled, they would work as they do now.

Adding particles to the tables would be cool anyway with or without this config option and would make it look cooler and a bit realistic

AlexIIL:
I'm not sure this will happen for the 7.1.x branch (its a simple-ish change, but its up to @asiekierka how it would work as I don't think 7.1.x wants any more items added), but this will happen for 8.0.x.

The robot goggles are still planned - they could be renamed "builders goggles" and be used for this though.

@lllMohandarlll can you rename this issue to something more descriptive than "Suggestion" please?

asiekierka:
Please do keep the lasers on as a default, by the way.

7.1.x probably won't get this - you'll have better luck patching it out yourself, shouldn't be too difficult.

AlexIIL:
Please do keep the lasers on as a default, by the way.

Defiantly.

AEnterprise:
@AlexIIL is this a better name? also you should be able to rename issues as well, there is an edit button next to the title

lllMohandarlll:
Meh..
I really hope this could be feasible for 1.7.x most of all... my pack is based on that version (as many other packs out there), since there are still plenty of mods and contents and tools in 1.7.x and these allows a complete and viable customization.
Overall, terrafirmacraft (the base of the pack I'm working on) and galacticraft are still for 1.7.10 only, unluckly, so I really could not move on
[...]

PR #3788, opened 2017-08-24
Invisible Lasers

JustWannaFly:
[Invisible Lasers](https://github.com/BuildCraft/BuildCraft/issues/3513) implemented via configuration option. Currently in the experimental configuration category, but that is subject to change once a proper place is figured out and the ability to view invisible lasers with goggles is implemented

### DIFF
--- a/common/buildcraft/silicon/BCSilicon.java
+++ b/common/buildcraft/silicon/BCSilicon.java
@@ -8,7 +8,6 @@ package buildcraft.silicon;
 
 import java.util.function.Consumer;
 
-import buildcraft.lib.config.EnumRestartRequirement;
 import net.minecraftforge.fml.common.Mod;
 import net.minecraftforge.fml.common.event.FMLInitializationEvent;
 import net.minecraftforge.fml.common.event.FMLPostInitializationEvent;
@@ -44,9 +43,6 @@ public class BCSilicon {
         BCSiliconConfig.preInit();
         BCSiliconItems.preInit();
         BCSiliconBlocks.preInit();
-
-        // Reload after all of the definitions have been created.
-        BCSiliconConfig.reloadConfig(EnumRestartRequirement.GAME);
 
         NetworkRegistry.INSTANCE.registerGuiHandler(INSTANCE, BCSiliconProxy.getProxy());
 
--- a/common/buildcraft/silicon/BCSilicon.java
+++ b/common/buildcraft/silicon/BCSilicon.java
@@ -8,6 +8,7 @@ package buildcraft.silicon;
 
 import java.util.function.Consumer;
 
+import buildcraft.lib.config.EnumRestartRequirement;
 import net.minecraftforge.fml.common.Mod;
 import net.minecraftforge.fml.common.event.FMLInitializationEvent;
 import net.minecraftforge.fml.common.event.FMLPostInitializationEvent;
@@ -40,8 +41,12 @@ public class BCSilicon {
     public void preInit(FMLPreInitializationEvent evt) {
         RegistryHelper.useOtherModConfigFor(MODID, BCCore.MODID);
 
+        BCSiliconConfig.preInit();
         BCSiliconItems.preInit();
         BCSiliconBlocks.preInit();
+
+        // Reload after all of the definitions have been created.
+        BCSiliconConfig.reloadConfig(EnumRestartRequirement.GAME);
 
         NetworkRegistry.INSTANCE.registerGuiHandler(INSTANCE, BCSiliconProxy.getProxy());
 
--- a/common/buildcraft/silicon/BCSiliconConfig.java
+++ b/common/buildcraft/silicon/BCSiliconConfig.java
@@ -6,14 +6,17 @@
 
 package buildcraft.silicon;
 
-import buildcraft.api.BCModules;
-import buildcraft.core.BCCoreConfig;
-import buildcraft.lib.config.EnumRestartRequirement;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.common.config.Configuration;
 import net.minecraftforge.common.config.Property;
 import net.minecraftforge.fml.client.event.ConfigChangedEvent.OnConfigChangedEvent;
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
+
+import buildcraft.api.BCModules;
+
+import buildcraft.lib.config.EnumRestartRequirement;
+
+import buildcraft.core.BCCoreConfig;
 
 public class BCSiliconConfig {
 
--- a/common/buildcraft/silicon/BCSiliconConfig.java
+++ b/common/buildcraft/silicon/BCSiliconConfig.java
@@ -28,16 +28,14 @@ public class BCSiliconConfig {
         Configuration config = BCCoreConfig.config;
         //TODO change the category and update the comment once visible with goggles is implemented
         propRenderLaserBeams = config.get("experimental", "renderLaserBeams", true,
-                "When set to false Laser beams will not be visible. Default: true");
+                "When false laser beams will not be visible while transmitting power");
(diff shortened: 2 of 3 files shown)


AEnterprise:
travis failure is again travis killing the build for some reason, gona have to investigate it
you might also want to link that email to your github (or use your github email for future commits)

common/buildcraft/silicon/BCSiliconConfig.java
```diff
@@ -0,0 +1,52 @@
...
+        Configuration config = BCCoreConfig.config;
+        //TODO change the category and update the comment once visible with goggles is implemented
+        propRenderLaserBeams = config.get("experimental", "renderLaserBeams", true,
+                "When false laser beams will not be visible while transmitting power");
+        EnumRestartRequirement.WORLD.setTo(propRenderLaserBeams);
+
+        MinecraftForge.EVENT_BUS.register(BCSiliconConfig.class);
+    }
+
+    public static void reloadConfig(EnumRestartRequirement restarted) {
```
AEnterprise: the restarted param is not needed, it's not used at all

common/buildcraft/silicon/BCSiliconConfig.java
```diff
@@ -0,0 +1,52 @@
...
+        renderLaserBeams = propRenderLaserBeams.getBoolean();
+    }
+
+    @SubscribeEvent
+    public static void onConfigChange(OnConfigChangedEvent cce) {
+        if (BCModules.isBcMod(cce.getModID())) {
+            EnumRestartRequirement req = EnumRestartRequirement.NONE;
```
AEnterprise: the restart requirement should always be NONE, it is applied instantly and we don't need a world reload to start/stop rendering

common/buildcraft/silicon/BCSiliconConfig.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static Property propRenderLaserBeams;
+
+    public static void preInit() {
+
+        Configuration config = BCCoreConfig.config;
+        //TODO change the category and update the comment once visible with goggles is implemented
+        propRenderLaserBeams = config.get("experimental", "renderLaserBeams", true,
+                "When false laser beams will not be visible while transmitting power");
+        EnumRestartRequirement.WORLD.setTo(propRenderLaserBeams);
```
AEnterprise: See below, the requirement should always be NONE, just set it here and leave it at none then

common/buildcraft/silicon/BCSilicon.java
```diff
@@ -40,9 +41,13 @@
     public void preInit(FMLPreInitializationEvent evt) {
         RegistryHelper.useOtherModConfigFor(MODID, BCCore.MODID);
 
+        BCSiliconConfig.preInit();
         BCSiliconItems.preInit();
         BCSiliconBlocks.preInit();
 
+        // Reload after all of the definitions have been created.
+        BCSiliconConfig.reloadConfig(EnumRestartRequirement.GAME);
```
AlexIIL: Put this in the config preinit please

common/buildcraft/silicon/BCSiliconConfig.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraftforge.common.config.Configuration;
+import net.minecraftforge.common.config.Property;
+import net.minecraftforge.fml.client.event.ConfigChangedEvent.OnConfigChangedEvent;
+import net.minecraftforge.fml.common.Loader;
+import net.minecraftforge.fml.common.LoaderState;
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
+
+public class BCSiliconConfig {
+
+    public static Boolean renderLaserBeams = true;
```
AlexIIL: Bit minor, but this should be a primitive, not the object type.

common/buildcraft/silicon/BCSiliconConfig.java
AlexIIL: You can just remove this line to make it work.

AlexIIL:
Thanks! As for goggles, the item can be found in BCCoreItems.goggles, and you can probably add a check against Minecraft.gteMinecraft().getPlayer(), and see if the item equipped on their head equals the goggles.

AEnterprise:
@AlexIIL the restart params sound like something that isn't needed nor usefull at all, they are only used to tell forge what needs to be restarted, it shoudn't be used at all to tell a config what reloaded, they shoudn't care about that at all

common/buildcraft/silicon/BCSiliconConfig.java
JustWannaFly: Apparently I've been doing too much JavaScript at work lately if I can't remember how to properly initialize a typed variable :stuck_out_tongue_winking_eye:

AlexIIL:
@AEnterprise the reload parameter is used both to tell forges gui what should be enabled/disabled in the gui, and so that Buildcraft only changes the values that can actually be changed during the cycle.

@JustWannaFly please leave the reload parameter as-is, it's simpler if all config should have the same method signature (as then we can move the onConfig changed event to lib and add the methods to a single list for reloading.

AEnterprise:
@AlexIIL that is not how configs or properties work at all

if we set a property to require a world restart then that property cannot be eddited ingame, you need to leave the world and edit from the main menu to be able to edit those. So they can safely be updated as the only time they are actually changed is when it's safe to edit those

as for configs requiring a minecraft restart those are also not editable ingame, excluding those from being reloaded also doesn't make sense because those are values we only need during startup so changing them doesn't hurt anything (and forge makes the player click 'i understand' about the needing a restart)

JustWannaFly:
I'm seeing the build fail with the below message:

:extractRangemapReplacedMain
/home/travis/build/BuildCraft/BuildCraft/build/sources/main/java
/home/travis/.travis/job_stages: line 57:  3702 Killed                  ./gradlew build
The command "./gradlew build" exited with 137.
Done. Your build exited with 1.

Any insight into what causes this? My googling just reveals potential issues with running out of memory or needing the travis set up in a container. I admit I don't know anything about travis though.

common/buildcraft/silicon/client/render/RenderLaser.java
```diff
@@ -57,4 +59,15 @@ public void renderTileEntityFast(@Nonnull TileLaser tile, double x, double y, do
             Minecraft.getMinecraft().mcProfiler.endSection();
         }
     }
+
+    private boolean isPlayerWearingBCGoggles () {
```
AlexIIL: Just `isPlayerWearingGoggles`, or lowercase `Bc`

common/buildcraft/silicon/BCSiliconConfig.java
```diff
@@ -24,9 +24,8 @@
     public static void preInit() {
 
         Configuration config = BCCoreConfig.config;
-        //TODO update the comment once visible with goggles is implemented
         propRenderLaserBeams = config.get("display", "renderLaserBeams", true,
-                "When false laser beams will not be visible while transmitting power");
+                "When false laser beams will not be visible while transmitting power without wearing the BC Goggles");
```
AlexIIL: Please use the full name (`BuildCraft`) or don't, rather than `BC`

AlexIIL:
Ok, the memory issue is fixed on the 8.0.x branch, so can you merge and push?

JustWannaFly:
Switched to just checking the head slot instead of iterating through all the armor slots. Definitely cleaner, but this is the first mod I've worked on so please keep letting me know if I'm doing anything silly that can be done a lot simpler.

JustWannaFly:
My apologies for the multiple failed builds. It took me a bit to get Intellij configured properly. All should be good now

AlexIIL:
Looks good, thanks :)

I think there are a few things that need changing about the config system, but some other classes want changing as well so I'll fix them later.

PR #3796, opened 2017-08-29
Laser networking optimizations

JustWannaFly:
- Laser only sends packet on target change 
- Laser beam position calculated on client side

### DIFF
--- a/common/buildcraft/silicon/tile/TileLaser.java
+++ b/common/buildcraft/silicon/tile/TileLaser.java
@@ -48,6 +48,7 @@ import buildcraft.silicon.client.render.AdvDebuggerLaser;
 public class TileLaser extends TileBC_Neptune implements ITickable, IDebuggable {
     private int ticks = 0;
     private BlockPos targetPos;
+    private BlockPos previousTarget;
     private final AverageLong avgPower = new AverageLong(100);
     private long averageClient;
     private final MjBattery battery;
@@ -107,7 +108,7 @@ public class TileLaser extends TileBC_Neptune implements ITickable, IDebuggable 
     }
 
     private void updateLaser() {
-        if (getTarget() != null) {
+        if (targetPos != null) {
             laserPos = new Vec3d(targetPos)
                 .addVector(
                     (5 + world.rand.nextInt(6) + 0.5) / 16D,
@@ -116,6 +117,20 @@ public class TileLaser extends TileBC_Neptune implements ITickable, IDebuggable 
                 );
         } else {
             laserPos = null;
+        }
+    }
+
+    private boolean changedTarget() {
+        if (previousTarget == null && targetPos != null) {
+            return true;
+        }
+        if (previousTarget != null && targetPos == null) {
+            return true;
+        }
+        if (previousTarget != null) {
+            return !previousTarget.equals(targetPos);
+        } else {
+            return false;
         }
     }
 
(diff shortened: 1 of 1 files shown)


common/buildcraft/silicon/tile/TileLaser.java
```diff
@@ -119,6 +120,20 @@ private void updateLaser() {
...
+        }
+        if (previousTarget != null && targetPos == null) {
+            return true;
+        }
+        if (previousTarget != null) {
+            return !previousTarget.equals(targetPos);
+        } else {
+            return false;
+        }
+    }
```
AlexIIL: You can use ````! Objects.equal(previousTarget, targetPos) ````

common/buildcraft/silicon/tile/TileLaser.java
```diff
@@ -129,41 +144,46 @@ public long getMaxPowerPerTick() {
...
-            return;
-        }
-        avgPower.tick();
         ticks++;
+        if (world.isRemote) {
+            // set laser render position on client side
+            if (ticks % (5 + world.rand.nextInt(10)) == 0 || targetPos == null) {
+                updateLaser();
+            }
+        } else {
```
AlexIIL: Please keep the old ````return```` statement rather than this else. Unless you have a reason for it?

common/buildcraft/silicon/tile/TileLaser.java
JustWannaFly: Yeah, that's much better. Being rusty in my Java caught me again. I'll get that fixed

common/buildcraft/silicon/tile/TileLaser.java
JustWannaFly: I just thought the if/else structure was cleaner for showing the client/server side updates. I can switch it back to having the old return if that's the pattern the buildcraft code uses.

AlexIIL: I mostly use
````java
 if (client) {
    // Do client tasks
    return;
}
````
as it makes all of the next lines less indented than otherwise (most of the time this doesn't matter, but sometimes it allows much longer lines to fit)

File: common/buildcraft/silicon/tile/TileLaser.java [hunk 3]
```diff
@@ -129,11 +130,18 @@ public long getMaxPowerPerTick() {
...
+            if (ticks % (5 + world.rand.nextInt(10)) == 0 || targetPos == null) {
+                updateLaser();
+            }
             return;
         }
+        // set target tile on server side
         avgPower.tick();
-        ticks++;
+
+        BlockPos previousTargetPos = targetPos != null ? targetPos.toImmutable() : null;
```
afdw: So now `previousTargetPos` is used only once. I think it could be inlined, an making it immutable is useless here (as you are using it only for comparing, and because this method only makes sense in `MutableBlockPos`, in normal `BlockPos`es it's just `return this;`).

JustWannaFly: My reason for using `toImmutable` was to get a copy of the target position without pointing a new reference to the same `BlockPos` in memory. Is there a better way to get a copy of a `BlockPos`?

AlexIIL: BlockPos is immutable, so there's very little point in making a copy. When you change targetPos later, it creates a new, different block pos.

JustWannaFly:
Apologizes for how slow this simple change was in coming. Life's been life again and kept me from any coding that isn't for work lately

common/buildcraft/silicon/tile/TileLaser.java
```diff
@@ -129,11 +130,18 @@ public long getMaxPowerPerTick() {
 
     @Override
     public void update() {
+        ticks++;
         if (world.isRemote) {
+            // set laser render position on client side
+            if (ticks % (5 + world.rand.nextInt(10)) == 0 || targetPos == null) {
```
AlexIIL: Replace this with a field ````private final SafeTimeTracker clientLaserMoveInterval = new SafeTimeTracker(5, 10);```` and ````if (clientLaserMoveInterval.markTimeIfDelay(world) || targetPos == null) {````. This also means that you can remove the ticks field. (Currently world.rand.nextInt(10) will return a different value every time, so this might take a lot longer than 5 - 15 ticks to change, which is a bit confusing).

JustWannaFly: Should I do the same for the server side targeting to keep things consistent?

AlexIIL: Yes please :)

AlexIIL:
Thanks :)

PR #3805, opened 2017-09-10
Fix quarry behavior, fixes #3740, fixes #3765 and fixes #3776

### LINKED ISSUE #3740: Quarry cannot handle non breakable fluids
I was using a quarry when it ran into some oil from galacticraft.  It then froze, and glitched back up to the top of the last fluid it hit. Please fix this.

### LINKED ISSUE #3765: Quarry stops before bedrock by 8 blocks.
Title says it. I just got a new extreme reactor. and then i gone to my quarry to find that it finished but it didnt reach bedrock! ( i knew it finished because any type of engine isn't working also)
my set up : http://i.imgur.com/PLYqTRW.png

### LINKED ISSUE #3776: 1.11.2 - quarry erroneously waste a lot of time and energy after solid block reappears in the mining area
**Buildcraft** version: 7.99.6
**Forge** version: 1.11.2-13.20.1.2429

In case a solid block is placed in the area that the quarry mined already (a few layers above for example), after it's demolition the mining tip does not go to the next solid block available using the shortest path - but instead it goes sequentially all through the empty areas, block by block and layer by layer.

iTitus:
Video with new behavior [here](https://youtu.be/dV9wdlfS18c)

### DIFF
--- a/common/buildcraft/builders/tile/TileQuarry.java
+++ b/common/buildcraft/builders/tile/TileQuarry.java
@@ -968,9 +968,9 @@ public class TileQuarry extends TileBC_Neptune implements ITickable, IDebuggable
         @Override
         public boolean equals(Object o) {
             return this == o ||
-                    !(o == null || getClass() != o.getClass()) &&
-                            from.equals(((TaskMoveDrill) o).from) &&
-                            to.equals(((TaskMoveDrill) o).to);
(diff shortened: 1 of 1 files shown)


AlexIIL:
Looks good, thanks :)

iTitus:
Nice, always happy to help ;)

PR #3819, opened 2017-09-21
Fix message names, fix IDebuggables threads and sides

afdw:


common/buildcraft/core/client/RenderTickListener.java
```diff
@@ -106,56 +110,33 @@ public static void renderOverlay(RenderGameOverlayEvent.Text event) {
...
-            appendDiff(right, serverRight, clientRight, headerFirst, headerSecond);
-
-        }
+            appendDiff(left, ClientDebuggables.SERVER_LEFT, clientLeft, headerFirst, headerSecond);
+            appendDiff(right, ClientDebuggables.SERVER_RIGHT, clientRight, headerFirst, headerSecond);
+        });
     }
 
-    private static IDebuggable getDebuggableObject(RayTraceResult mouseOver) {
+    private static <T extends TileBC_Neptune & IDebuggable> Optional<T> getDebuggableObject(RayTraceResult mouseOver) {
```
AlexIIL: Don't mandate TileBC_Neptune (was that a typo? You never checked for TileBC_Neptune in an instanceof check). In the future this might also be entities as well.

common/buildcraft/core/client/RenderTickListener.java
```diff
@@ -106,56 +110,33 @@ public static void renderOverlay(RenderGameOverlayEvent.Text event) {
...
-        EnumFacing side = mouseOver.sideHit;
-        if (server == null) {
-            client.getDebugInfo(left, right, side);
-        } else {
-            List<String> serverLeft = new ArrayList<>();
-            List<String> serverRight = new ArrayList<>();
 
+        getDebuggableObject(mouseOver).ifPresent(client -> {
+            EnumFacing side = mouseOver.sideHit;
+            MessageManager.sendToServer(new MessageDebuggableRequest(client.getPos(), side));
```
AlexIIL: This sends a message to the server, every frame. And on the server it gets debugging information, and sends it back -- note that some clients might have upwards of 120 fps, which is over 6 times the absolute maximum that the information might change (20 times a second).

common/buildcraft/core/client/RenderTickListener.java
```diff
@@ -106,56 +110,33 @@ public static void renderOverlay(RenderGameOverlayEvent.Text event) {
...
+            client.getClientDebugInfo(clientLeft, clientRight, side);
 
             final String headerFirst = DIFF_HEADER_FORMATTING + "SERVER:";
             final String headerSecond = DIFF_HEADER_FORMATTING + "CLIENT:";
-            appendDiff(left, serverLeft, clientLeft, headerFirst, headerSecond);
-            appendDiff(right, serverRight, clientRight, headerFirst, headerSecond);
-
-        }
+            appendDiff(left, ClientDebuggables.SERVER_LEFT, clientLeft, headerFirst, headerSecond);
+            appendDiff(right, ClientDebuggables.SERVER_RIGHT, clientRight, headerFirst, headerSecond);
```
AlexIIL: Can you change ````IDebuggable.getClientDebugInfo```` to do nothing by default? That way you can do diffs for the server-sided debug infos only, and then just append the client data afterwards. (Currently the diff algorithm only emits red exclamation marks if the server + client lists are the same length)

afdw: The problem is that some tiles had different server and client infos, not server and some additional lines on client.

common/buildcraft/core/client/RenderTickListener.java
afdw: I rewrote code a few times, one of implementations was inside TileBC_Neptune.

afdw:
@AlexIIL OK, fixed.

PR #3829, opened 2017-10-17
For 1.10.2 Request

dagantang:
Could you consider the version 1.10.2 again?
I see your MOD very interested
But I am playing the version is 1.10.2
And there are a lot of MOD are still 1.10.2
So ask you to think about it again
Perhaps in 1.12 after the production of 1.10.2 version
Thank you very much for seeing my reply


AlexIIL:
I'm not interested in putting in the time necessary to backport buildcraft to 1.10. (And from what I've heard from the rest of the team, no-one else is either). I'm hoping that when we forward port to 1.12 a lot of other mods will also target that version, so this won't be much of an issue any more.

(Note: you should have been an issue rather than a pull request, although it won't change the response)

Issue #3884, opened 2017-11-30
Powering an Obsidian Pipe with no connections causes crash


RhysWhy:
**Buildcraft** version: Currently the one in use on the dev branch
**Forge** version: Currently the one in use on the dev branch
Link to crashreport or log (on pastebin or gist, please don't put the entire log here) : 
Singleplayer/multiplayer: Singleplayer (multi-untested)

Additional information: Reported on the discord by another user, powering an obsidian pipe with a creative engine causes a null reference when the pipe has no connections, as there is no 'sucking direction'.

I am able to reproduce this reliably and am working on a fix.

PR #3885, opened 2017-11-30
Obsidian pipes no longer draw power when not connected to other pipes

RhysWhy:
As per #3884 obsidian pipes that were not connected to any other pipes caused a crash as they had nowhere to send items to.

To fix this, if there are no pipes to send items to then the obsidian pipes don't try to suck items and don't draw power (as they cannot function)

### DIFF
--- a/common/buildcraft/transport/pipe/behaviour/PipeBehaviourObsidian.java
+++ b/common/buildcraft/transport/pipe/behaviour/PipeBehaviourObsidian.java
@@ -191,18 +191,22 @@ public class PipeBehaviourObsidian extends PipeBehaviour implements IMjRedstoneR
             return microJoules;
         }
         EnumFacing openFace = getOpenFace();
-        for (int d = 1; d < 5; d++) {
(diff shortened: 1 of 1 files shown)


common/buildcraft/transport/pipe/behaviour/PipeBehaviourObsidian.java
```diff
@@ -191,18 +191,22 @@ public long receivePower(long microJoules, boolean simulate) {
             return microJoules;
         }
         EnumFacing openFace = getOpenFace();
-        for (int d = 1; d < 5; d++) {
```
AlexIIL: It would probably be cleaner to do the same as the above if statement: do "if openFace == null return microJoules" rather than wrapping everything in an additional indent level.

RhysWhy:
updated to check face prior to for loop, rather than wrapping around the for loop

AlexIIL:
Thanks. Looks good now :)
(apart from minor that's-not-quite-how-I-format-code, but that's too minor.)

PR #3929, opened 2017-12-21
Store tank capacities in nbt data

Indemnity83:
This is related to our discussions in Discord about reloading tank tile entities with capacities other than the default 16 buckets. I see three possible solutions to this issue: 

1) Addons like IronTanks implement their own tile (which can still extend the BC TileTank) which has the correct magic number of buckets in the constructor (or as a static property if the most recent commit is rolled back):

```
public class CustomTank extends TileTank {
    public TileTank() {
        this(45 * Fluid.BUCKET_VOLUME);
    }
} 
```

2) Implement reading/writing the tank capacity from/to NBT data in the BuildCraft tank (**this PR**). 

3) Push this functionality upstream into the Forge FluidTank class (not sure what the other implications are of this though).


AlexIIL:
You need to do this in your own mod, and (most likely) validate the read capacity or store it by type.

PR #4024, opened 2018-02-12
Allow tanks to be explicitly associated

Indemnity83:
Tanks are currently implicitly associated when they're stacked together. This PR allows tanks to also be explicitly associated through an association key provided on the tile. 

The purpose of this is to allow third party add-ons (like Iron Tanks) to explicitly define and track an association key for their own tanks and control what tanks can pass fluid down (or up) the stack. 

Current implementation leaves association key for all Buildcraft tanks set to NULL, so no change in current behavior, any stacked set of tanks will allow fluid/gas flow between the tanks in the stack. 

Specific use case is for Iron Tanks Creative and Void tanks which should not stack with any other tanks. 

Potential future use case could include stained glass tanks which don't allow fluid flow between different colored tanks.

### DIFF
--- a/common/buildcraft/factory/tile/TileTank.java
+++ b/common/buildcraft/factory/tile/TileTank.java
@@ -193,7 +193,7 @@ public class TileTank extends TileBC_Neptune implements ITickable, IDebuggable, 
         TileEntity tile = world.getTileEntity(at);
         if (tile instanceof TileTank) {
             TileTank tileTank = (TileTank) tile;
-            if (tileTank.canConnectTo(this)) {
+            if (tileTank.canConnectTo(this) && this.canConnectTo(tileTank)) {
                 return tileTank;
             }
         }
--- a/common/buildcraft/factory/tile/TileTank.java
+++ b/common/buildcraft/factory/tile/TileTank.java
@@ -10,7 +10,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 
 import net.minecraft.entity.EntityLivingBase;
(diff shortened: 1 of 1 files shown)


afdw:
Hm, `boolean canConnectTo(TileTank tile)` might be better I think? To add ability to implement any needed behavior and not to add field that isn't used by BC.

Indemnity83:
👍  that cleans up the PR footprint a bit too; you'll probably want to squash and merge, I left the commit history just for clarity of the discussion.

afdw:
And don't we need to check both (`tileTank.canConnectTo(this) && this.canConnectTo(tileTank)`)?

AlexIIL:
Perhaps also add an EnumFacing direction argument? I know that it can be computed easily anyway, but it might be useful for future additions.

Although this is fairly minor, it doesn't really matter for this PR right now, as I can always clean it up after merging.

Indemnity83:
Adding EnumFacing will require walking the stack recursively instead of just looping. Not sure if you want me to try to do that here, or maybe we just add it as a separate PR/commit?

Really, the suggestion by @afdw should be walked recursively too, but I think we can get by without that and still have predictable behavior.

AlexIIL:
I think this is a simple enough use case for me to do it myself. This looks good as-is anyway, as the EnumFacing is technically just extra info, so I'll merge it and play around with alternatives for a bit.

AlexIIL:
Would the changes in https://github.com/BuildCraft/BuildCraft/commit/2d74d9352a4bd3c3d994d5e95d59741d559613c8 be acceptable @Indemnity83 ?

Indemnity83:
I like it; much easier than the recursion I was thinking was necessary.

PR #4029, opened 2018-02-15
Extract a public reflow method for balancing tank capacities

Indemnity83:
If the amount of fluid is directly adjusted on a single tile's tank instance the tank stack doesn't update. Buildcraft doesn't directly manipulate single tile tank instances, but add-ons can (in particular see indemnity83/irontanks#57). 

This PR simply pulls the logic used when a new tank block is added to a stack out into a public method that can be called to force a reflow of the tanks fluid/gas. 

I'm not totally clear on the purpose of `isPlayerInteracting`; it may be that toggling this during reflow needs to be passed as a parameter: `public void reflowTanks(boolean doPlayerInteraction)` or similar.

### DIFF
--- a/common/buildcraft/factory/tile/TileTank.java
+++ b/common/buildcraft/factory/tile/TileTank.java
@@ -104,34 +104,7 @@ public class TileTank extends TileBC_Neptune implements ITickable, IDebuggable, 
     public void onPlacedBy(EntityLivingBase placer, ItemStack stack) {
         super.onPlacedBy(placer, stack);
         if (!placer.world.isRemote) {
-            List<TileTank> tanks = getTanks();
-            FluidStack fluid = null;
-            for (TileTank tile : tanks) {
-                FluidStack held = tile.tank.getFluid();
-                if (held == null) {
-                    continue;
-                }
-                if (fluid == null) {
-                    fluid = held;
-                } else if (!fluid.isFluidEqual(held)) {
-                    return;
-                }
-            }
-            if (fluid == null) {
-                return;
-            }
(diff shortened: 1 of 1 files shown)


AlexIIL:
This looks good, however it's simple enough for me to just extract a method manually.

````isPlayerInteracting```` is used for checking to see if we need to speed up the flow animation in a tank (as otherwise right clicking with a bucket takes ages to complete properly).

I'm not sure this is strictly necessary however, as the void tank could do this:
````java
// We specifically only want to drain fluid from *this*
// tank; not the whole tank stack. This lets a user put the
// void tank at the top of a stack as overflow prevention
if (this.tank.getFluidAmount() > 0) {
     this.drain(Math.min(this.tank.getFluidAmount(), transferPerTick), true);
}
````

Indemnity83:
Thanks AlexIIL; 

I can't use `this.drain` because that would cause the void tank to void all fluid in a tank stack. The intended behavior is to only void fluid that is in the void tank itself. 

So if you place a void tank on top of a regular tank. You should be able to fill the bottom tank to 16 buckets no problem, but anything beyond 16 buckets would end up in the void tank and be voided.

AlexIIL:
Yes, hence the check for ````this.tank.getFluidAmount()````. (As ````this.tank```` will only contain fluid if the fluid has filled up to the void tank. At least, that's the theory, although I've never tested it)

AlexIIL:
Oh, you've already implemented it. I guess that means it worked?

Indemnity83:
Seemed to work in my test just now.

PR #4032, opened 2018-02-17
Fix fluid rendering for tanks which cannot stack

### LINKED ISSUE #4031: [7.99.15] Fluid render bug when stacking non-connectable tanks
BuildCraft version: 7.99.15
Forge version: 14.23.2.2611
Link to crash report or log: N/A
Singleplayer or multiplayer: N/A
Additional information: 

When placing a non-connecting tank (using canConnectTo() public method) on top of any other tank; if there is any fluid in the upper tank, the lower tank does not render the top face of the fluid.

Indemnity83:
Added a check when determining `connectedUp` & `connectedDown` to ensure that tanks are only 'connectedUp/Down' when the tanks being checked are allowed to be connected. 

fixes #4031


AlexIIL:
I'm going to assume you've already tested this yourself, as buildcraft doesn't use this itself :p

Indemnity83:
Tested by temporarily setting the native buildcraft tank's 'canConnectTo()' to false and confirming fluid rendering :)

PR #4118, opened 2018-04-06
Finally fixed #3503

### LINKED ISSUE #3503: BC 7.1.20 AutoWorkbench Dupe (NBT checking)
AutoWorkbench does not compare NBT items.
https://www.youtube.com/watch?v=rfqad_-kMmY&t=2m26s

I fixed it activating the check here: https://github.com/BuildCraft/BuildCraft/blob/7.1.x/common/buildcraft/factory/TileAutoWorkbench.java#L340

Why don't you use **ItemStack.areItemStackTagsEqual()** in here?
https://github.com/BuildCraft/BuildCraft/blob/7.1.x/common/buildcraft/core/lib/inventory/StackHelper.java#L205-L207
If the first item is missing NBT, it will return true, which is not good.

usernameak:
Now it really isn't dupeable.

### DIFF
--- a/common/buildcraft/factory/TileAutoWorkbench.java
+++ b/common/buildcraft/factory/TileAutoWorkbench.java
@@ -359,7 +359,7 @@ public class TileAutoWorkbench extends TileBuildCraft implements ISidedInventory
 
 					ItemStack inputInvStack = inputInv.getStackInSlot(j);
 
-					if (StackHelper.isMatchingItem(inputInvStack, comparedStack, true, false)
+					if (StackHelper.isMatchingItem(inputInvStack, comparedStack, true, true)
 							&& inputInvStack.stackSize > bindingCounts[j]) {
 						found = true;
 						bindings[i] = j;


AlexIIL:
I haven't booted into a 1.7.10 workspace for quite a while... so it may take a while for me to validate this (sorry)

AlexIIL:
That being said I'm really not sure how also validating NBT could be harmful...

asiekierka:
Leave this for review for me. I'm unable to do it today, I'll do it tomorrow.

asiekierka:
I *think* there is a reason I made it that way. Will have to investigate as it was a long, long time ago.

AEnterprise:
travis build is failing but is unrelated to the PR changes, somehow the JDK is choking internaly on establishing an https connection to pull in gradle

asiekierka:
@usernameak Can you provide a reproduction scenario?

usernameak:
@asiekierka look at the video in #3503 . Note that you have to split cobble stack in two otherwise it won't work.

asiekierka:
Checked the video. The issue shown in the video is not reproducible in 7.1.23 - the auto workbench still operates, but the recipe output is correct for the *supplied* items, not the fake "configured" items.

PR #4127, opened 2018-04-15
Fix #4082 - Fluids can no longer be broken by Mining Well

### LINKED ISSUE #4082: Mining Well can break oil & fuel
BuildCraft version: 7.99.15
Forge version: 14.23.2.2611
Link to crash report or log:  None
Singleplayer or multiplayer:  Both
Steps to reproduce: Mining Well
Additional information: 

https://www.youtube.com/watch?v=PEBGvnFUlL8
javaw 14-03-2018 17-04-20-745
-Can not break Water and Lava

LordGolias:
The code follows the same pattern as the code for the Quarry.

I tested this in-game for both lava, a fuel, and no liquid, confirming expected behavior (stops when a fluid is present)

### DIFF
--- a/buildcraft_resources/changelog/7.99.17
+++ b/buildcraft_resources/changelog/7.99.17
@@ -2,3 +2,4 @@
 Bug fixes:
 
 * [#4124] The direction of an iron pipe cannot be changed by clicking the centre unless the pipe is already facing a direction.
+* [#4082] Fluids can no longer be mined by Mining Well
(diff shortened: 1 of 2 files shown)


AlexIIL:
This looks good - thanks!

Just a minor thing, it would be quite helpful to add an entry to https://github.com/BuildCraft/BuildCraft/blob/8.0.x/buildcraft_resources/changelog/7.99.17 with the issue number in brackets, and the issue title afterwards.

Also for future reference please don't replace old commits with new ones - it's much easier to track the changes if they appear independently. (In addition github has a "squash an merge" option for squashing all of them into one, so you don't need to do this yourself).

LordGolias:
Added entry to changelog. I agree with you. I am used to a different PR practice, but yeah, the practice is no longer relevant with the squash option. :) Thanks for pointing out.

AlexIIL:
Ok, thanks

Issue #3966, opened 2018-01-08
Add a config option to tweak the height of oil spouts.


AlexIIL:
Specifically an option for the maximum height that an oil spout can generate. While we're at it, why not also add options for a lot of the other hardcoded numbers? If we add a lot then it would be a good idea to add another section to ````worldgen```` - possibly
````
worldgen { 
    oil_spout {
        small_min_height
        small_max_height
        large_min_height
        large_max_height
    }
}
````

AlexIIL:
Implemented by @scagood in #4271, and will be a part of 7.99.19 at some point in the future.

PR #4130, opened 2018-04-17
Fix #4080 - Ported option to disable generation of oil spout and change rate of occurrence of large oil wells

### LINKED ISSUE #4080: Option to change the frequency of oil spouts
It would be nice if the frequency of these spouts could be adjusted (or even disabled without affecting pools).

LordGolias:
* Ported option to disable generation of oil spout.
* Ported option to change rate of occurrence of large oil wells.

These are literal ports of the old code. I do not like very much that "rate" is not really a rate of how often a oil well appears, but how likely it is for them to be larger. I think that that is fine as a port, but I definitely support a new issue to improve how oil is generated and what quantities of it should be part of the config, as per #3966. I haven't addressed #3966 because of that.

### DIFF
--- a/buildcraft_resources/changelog/7.99.17
+++ b/buildcraft_resources/changelog/7.99.17
@@ -1,7 +1,7 @@
 
 Ported content:
 
-* Re-added the config option to change the rate of oil well generation and generation of a spout.
+* [#4080] Re-added the config option to change the rate of oil well generation and generation of a spout.
 
 Bug fixes:
 
--- a/buildcraft_resources/changelog/7.99.17
+++ b/buildcraft_resources/changelog/7.99.17
@@ -1,4 +1,8 @@
 
+Ported content:
+
+* Re-added the config option to change the rate of oil well generation.
+
 Bug fixes:
 
 * [#4124] The direction of an iron pipe cannot be changed by clicking the centre unless the pipe is already facing a direction.
--- a/buildcraft_resources/changelog/7.99.17
+++ b/buildcraft_resources/changelog/7.99.17
@@ -1,7 +1,7 @@
 
 Ported content:
 
-* Re-added the config option to change the rate of oil well generation.
+* Re-added the config option to change the rate of oil well generation and generation of a spout.
 
 Bug fixes:
 
--- a/common/buildcraft/energy/generation/OilGenerator.java
+++ b/common/buildcraft/energy/generation/OilGenerator.java
@@ -104,6 +104,7 @@ public enum OilGenerator implements IWorldGenerator {
         boolean oilBiome = BCEnergyConfig.surfaceDepositBiomes.contains(biome.getRegistryName());
 
         double bonus = oilBiome ? 3.0 : 1.0;
+        bonus *= BCEnergyConfig.oilWellGenerationRate;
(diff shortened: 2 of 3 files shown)


common/buildcraft/energy/generation/OilGenerator.java
```diff
@@ -104,17 +104,21 @@ public void generate(Random random, int chunkX, int chunkZ, World world, IChunkG
...
+
+        if (random_number <= CHANCE_LARGE * bonus) {
             // 0.04%
             type = GenType.LARGE;
-        } else if (rand.nextDouble() <= CHANCE_MEDIUM * bonus) {
+        } else if (random_number <= CHANCE_MEDIUM * bonus) {
             // 0.1%
             type = GenType.MEDIUM;
-        } else if (oilBiome && rand.nextDouble() <= CHANCE_LAKE * bonus) {
+        } else if (oilBiome && random_number <= CHANCE_LAKE * bonus) {
```
AlexIIL: I wanted to change this originally, however this approach has a few problems:
 - We break existing worldgen. This will only be noticeable if an oil well goes across a chunk boundary, or if someone compares two oil spouts in different worlds. Obviously this was originally broken from 1.7.10 when I added the `MAGIC_GEN_NUMBER`, but I think it's more of an issue for newer worlds.
 - The chance constants (`CHANCE_LAKE`, `CHANCE_MEDIUM`, `CHANCE_LARGE`) need to be changed so that they also include the previously tested value. Something like this:
````java
     private static final double CHANCE_LARGE = 0.0004;
     private static final double CHANCE_MEDIUM = 0.001 + CHANCE_MEDIUM;
     private static final double CHANCE_LAKE = 0.02 + CHANCE_LAKE;
````
If/when these constants are available in the config it would probably be a good idea to do this addition automatically, and complain if the final number is greater than 1.

However I don't think the (tiny) performance gain is much of a gain compared to world/chunk breaks.

Comment (LordGolias) [reply common/buildcraft/energy/generation/OilGenerator.java]: Very good catch. I did not think of the update on the RNG state on each call (besides the bad math). Will revert it.

Comment (AlexIIL) [reply common/buildcraft/energy/generation/OilGenerator.java]: Yeah it's an annoying case of something that begs to be fixed but results in breaking compat if we do :/

**Comment by AlexIIL** [comment]
Sorry about checkstyle complaining, it looks like I didn't re-sort imports when I moved some of transport into silicon.

File: buildcraft_resources/changelog/7.99.17 [hunk 2]
```diff
@@ -1,4 +1,8 @@
 
+Ported content:
+
+* Re-added the config option to change the rate of oil well generation and generation of a spout.
```
AlexIIL: Oh, can you add `[#4080]` at the start?

AlexIIL:
Oh and just FYI github doesn't auto-link issues or PR's together when the *title* links to an issue number. So:

This fixes #4080.

common/buildcraft/energy/generation/OilGenerator.java
AlexIIL: Sorry for constantly changing my mind on this, but it's probably safer to leave it as-is (not adding a variable) and leaving it to be cleaned up later, probably alongside the additional config options.

AlexIIL:
Thanks :)

PR #4136, opened 2018-04-22
Improvements to Heat Exchanger code

LordGolias:
This allows the Heat Exchanger to receive multiples of 5, which allows to process every combination of fluids from the distiller.

Also removed the multiplier non-divisible by 1000 that was in contradiction to the comment.

af7743f is about allowing the HE to process multiple multiples. E.g. a 3-middle section HE can now process 10 units of liquid if it accepts it.

### DIFF
--- a/buildcraft_resources/changelog/7.99.17
+++ b/buildcraft_resources/changelog/7.99.17
@@ -1,4 +1,8 @@
 
+Improvements:
+
+* [#4135] Made Heat Exchanger process arbitrary amounts of fluid.
+
 Ported content:
 
 * [#4080] Re-added the config option to change the rate of oil well generation and generation of a spout.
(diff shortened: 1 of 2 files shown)


LordGolias:
I also prefer that. Let me get back to you ;)

AlexIIL:
Looks good, thanks!

Just a minor thing: mind adding an entry to the changelog? (Probably under a new "Improvements" section, like in the [7.99.16 changelog](https://github.com/BuildCraft/BuildCraft/blob/8.0.x/buildcraft_resources/changelog/7.99.16#L8-L10).

LordGolias:
Added. Thanks for the reminder.

AlexIIL:
Ok, looks perfect now.

PR #4164, opened 2018-05-13
Add missing vanilla rotation handlers, closes #4068

### LINKED ISSUE #4068: in 1.8.9 the wrench could be used to reposition stairs.
Any chance it could be added back in. Tried it in 1.11.2 as well and it didn't work there. 

thanks

BuildCraft version: 7.99.15
Forge version: 8.0.99.99
Link to crash report or log: 
Singleplayer or multiplayer: Single Player
Steps to reproduce: 
Additional information: tried it in MC 1.11.2 as well

iTitus:
I copied some vanilla logic to really find all valid rotations

### DIFF
--- a/common/buildcraft/lib/block/VanillaRotationHandlers.java
+++ b/common/buildcraft/lib/block/VanillaRotationHandlers.java
@@ -9,6 +9,7 @@ import java.util.function.Predicate;
 
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockAnvil;
+import net.minecraft.block.BlockBanner;
 import net.minecraft.block.BlockButton;
 import net.minecraft.block.BlockChest;
 import net.minecraft.block.BlockCocoa;
@@ -32,15 +33,24 @@ import net.minecraft.block.BlockPistonBase;
 import net.minecraft.block.BlockPumpkin;
 import net.minecraft.block.BlockRedstoneDiode;
 import net.minecraft.block.BlockShulkerBox;
+import net.minecraft.block.BlockSkull;
+import net.minecraft.block.BlockStairs;
+import net.minecraft.block.BlockStandingSign;
 import net.minecraft.block.BlockTorch;
+import net.minecraft.block.BlockTrapDoor;
 import net.minecraft.block.BlockTripWireHook;
+import net.minecraft.block.BlockWallSign;
 import net.minecraft.block.properties.IProperty;
 import net.minecraft.block.state.BlockFaceShape;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.tileentity.TileEntity;
+import net.minecraft.tileentity.TileEntitySkull;
 import net.minecraft.util.EnumActionResult;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
+
+import net.minecraftforge.fml.common.ObfuscationReflectionHelper;
 
 import buildcraft.api.blocks.CustomRotationHelper;
 import buildcraft.api.blocks.ICustomRotationHandler;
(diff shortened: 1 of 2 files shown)


common/buildcraft/lib/block/VanillaRotationHandlers.java
```diff
@@ -62,13 +80,27 @@
...
+        CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockAnvil.class, getHandlerHorizontalFreely(BlockAnvil.class));
+        CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockEnderChest.class, getHandlerHorizontalFreely(BlockEnderChest.class));
+        CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockFurnace.class, getHandlerHorizontalFreely(BlockFurnace.class));
+        CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockCocoa.class, VanillaRotationHandlers::rotateCocoa);
         CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockTorch.class, VanillaRotationHandlers::rotateTorch);
+        CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockLadder.class, VanillaRotationHandlers::rotateLadder);
+        CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockHopper.class, VanillaRotationHandlers::rotateHopper);
         CustomRotationHelper.INSTANCE.registerHandlerForAll(BlockChest.class, VanillaRotationHandlers::rotateChest);
+        // TODO: DoubleChest, BannerHanging, BannerStanding, Bed, DoublePlant, EndPortalFrame (?), Stairs, TrapDoor, WallSign, StandingSign, Skull, CommandBlock (?)
```
iTitus: Just leave a comment if I should add/remove something from this todo

iTitus:
Note: Not finished yet - waiting for feedback to continue

common/buildcraft/lib/block/VanillaRotationHandlers.java
AlexIIL: | BlockType | Coment |
|--|--|
| double chests | Hopefully this can just be alternated between the two directions, although I'm not sure if it needs anything else updating. |
| BlockBed | I'm not sure about beds, as respawning mechanics might require a specific part of the bed. |
| BlockEndPortalFrame | Should probably not be rotatable, as player's can't even break it in survival. Although it might be useful in creative? |
| BlockWallSign, BlockBannerHanging | Can you do the same thing with these as with ladders? (Although they might need different checks than ladders to match vanilla) | 
| BlockStandingSign, BlockBannerStanding  | In theory it should be possible to just increase the rotation angle by one on every rotation. |
| BlockStairs | I think it might be best to cycle round the facing property until we get to a predetermined one (say NORTH), then invert the `EnumHalf` bit so it flips from the top to the bottom instead. (And then it would have to flip on the previous one for bottom to top) |
| BlockTrapDoor | As vanilla doesn't doesn't care if a trap door doesn't face a particular side, we can probably do a similar thing to stairs. |
| BlockSkull | This is a bit trickier as the rotation data is either in the tile entity or in the block state. However it should be possible to do something with the skull - although perhaps treat the rotation types separately - never go from a wall-based skull to a ground-based one. |
| BlockCommandBlock | This might be useful, I'm not sure. This one is a bit sensitive due to command blocks being unforgiving to minor changes (it's basically code). |

AlexIIL: | BlockType | Coment |
|--|--|
| double chests | Hopefully this can just be alternated between the two directions, although I'm not sure if it needs anything else updating. |
| BlockBed | I'm not sure about beds, as respawning mechanics might require a specific part of the bed. |
| BlockEndPortalFrame | Should probably not be rotatable, as player's can't even break it in survival. Although it might be useful in creative? |
| BlockWallSign, BlockBannerHanging | Can you do the same thing with these as with ladders? (Although they might need different checks than ladders to match vanilla) | 
| BlockStandingSign, BlockBannerStanding  | In theory it should be possible to just increase the rotation angle by one on every rotation. |
| BlockStairs | I think it might be best to cycle round the facing property until we get to a predetermined one (say NORTH), then invert the `EnumHalf` bit so it flips from the top to the bottom instead. (And then it would have to flip on the previous one for bottom to top) |
| BlockTrapDoor | As vanilla doesn't doesn't care if a trap door doesn't face a particular side, we can probably do a similar thing to stairs. |
| BlockSkull | This is a bit trickier as the rotation data is either in the tile entity or in the block state. However it should be possible to do something with the skull - although perhaps treat the rotation types separately - never go from a wall-based skull to a ground-based one. |
| BlockCommandBlock | This might be useful, I'm not sure. This one is a bit sensitive due to command blocks being unforgiving to minor changes (it's basically code). |

iTitus:
I'll work on these in the coming days

AlexIIL:
Thanks!

iTitus:
Ready for review ;)

AlexIIL:
Well those "coming days" were a bit quick :p

common/buildcraft/lib/block/VanillaRotationHandlers.java
```diff
@@ -119,47 +172,212 @@ private static EnumActionResult rotateLever(World world, BlockPos pos, IBlockSta
...
+
+            return rotateOnce(world, pos, state, BlockTrapDoor.FACING, ROTATE_HORIZONTAL);
+        }
+        return EnumActionResult.PASS;
+    }
+
+    private static EnumActionResult rotateStairs(World world, BlockPos pos, IBlockState state, EnumFacing sideWrenched) {
+        if (state.getBlock() instanceof BlockStairs) {
+
+            if (sideWrenched.getAxis().isHorizontal()) {
```
AlexIIL: Please don't use this - it's a little confusing to use as getting the correct side every time can be annoying. Perhaps try with this instead? (Same applies to trapdoors, and with those the problem is even worse as all of the sides except the main face are small).
````java
private static EnumActionResult rotateStairs(World world, BlockPos pos, IBlockState state, EnumFacing sideWrenched) {
    if (state.getBlock() instanceof BlockStairs) {

        EnumFacing dir = state.getValue(BlockStairs.FACING);

        if (dir == ROTATE_HORIZONTAL.get(0)) {
            BlockStairs.EnumHalf half = state.getValue(BlockStairs.HALF);
            if (half == BlockStairs.EnumHalf.TOP) {
                half = BlockStairs.EnumHalf.BOTTOM;
            } else {
                half = BlockStairs.EnumHalf.TOP;
            }
            state = state.withProperty(BlockStairs.HALF, half);
        }

        return rotateOnce(world, pos, state, BlockStairs.FACING, ROTATE_HORIZONTAL);
    }
    return EnumActionResult.PASS;
}
````

iTitus: Same goes for the door tbh.

iTitus: But I will look in to it

AlexIIL:
Thanks! They seem to work well now.

PR #4192, opened 2018-06-09
changed getTankProperties to test for fluid on top

### LINKED ISSUE #4133: gates not recognizing gaseous fluid levels
I am trying to use gates to turn on a wire when the gaseous fuel in a tank reaches 25%. Regardless of the trigger I use the gate powers the wire unless the tank is full. My guess is I am not using the gate correctly or not using the right trigger. If this is a known issue I apologize for re-posting it.  BC .15 had the same behavior. 

MC 1.12.2, forge 14.23.1.2599, BC 7.99.16
BuildCraft version: 7.99.16
Forge version: 14.23.1.2599
Link to crash report or log: 
**Singleplayer** or multiplayer: 

Steps to reproduce: 
Additional information:

rosebeats:
the getTankProperties method previously did not account for gasseous liquids, which end up in the top tank.  The function now also checks the top tank in case there is gasseous liquid in it.  It appears this fixes #4133.

### DIFF
--- a/common/buildcraft/factory/tile/TileTank.java
+++ b/common/buildcraft/factory/tile/TileTank.java
@@ -262,7 +262,11 @@ public class TileTank extends TileBC_Neptune implements ITickable, IDebuggable, 
     public IFluidTankProperties[] getTankProperties() {
         List<TileTank> tanks = getTanks();
         TileTank bottom = tanks.get(0);
+        TileTank top = tanks.get(tanks.size() - 1);
         FluidStack total = bottom.tank.getFluid();
+        if(total == null) {
+            total = top.tank.getFluid();
+        }
         int capacity = 0;
         if (total == null) {
             for (TileTank t : tanks) {


asiekierka:
I don't like this solution. It should ,instead, check if the fluid is gaseous, not hack around like this.

rosebeats:
Yeah, that's why I did it that way.  It felt hacky to me too, but if it gets a liquid from the top tank, I figured it would be redundant to then check if it's gaseous.

AlexIIL:
Yes actually it looks like everywhere getTanks() is called is followed by `if (isGas) Collections.reverse(tanks), so moving those to the getTanks method would be sensible. However I'm happy to merge this as-is anyway, as it doesn't add worse hacky stuff than the already existing assumption.

asiekierka:
Code deduplication is always nice, so I'd do what you suggested if I were you, at least for now.

Issue #3610, opened 2017-05-24
Quarry crashes when reaching y=0 with no bedrock there


Barteks2x:
Exception from dev environment with version from curse 7.99.3:
```
net.minecraft.util.ReportedException: Ticking block entity
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:801) ~[MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:699) ~[MinecraftServer.class:?]
	at net.minecraft.server.integrated.IntegratedServer.tick(IntegratedServer.java:156) ~[IntegratedServer.class:?]
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:548) [MinecraftServer.class:?]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_131]
Caused by: java.lang.NullPointerException
	at net.minecraft.util.math.Vec3d.<init>(Vec3d.java:38) ~[Vec3d.class:?]
	at buildcraft.builders.tile.TileQuarry.update(TileQuarry.java:336) ~[TileQuarry.class:?]
	at net.minecraft.world.World.updateEntities(World.java:1957) ~[World.class:?]
	at net.minecraft.world.WorldServer.updateEntities(WorldServer.java:651) ~[WorldServer.class:?]
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:795) ~[MinecraftServer.class:?]
	... 4 more
```
(found when testing cubic chunks compatibility, but also happens with vanilla+forge+buildcraft).

The line of code in TileQuarry https://github.com/BuildCraft/BuildCraft/blob/c47e88531f0160ccc3c6e03720e5047f111a658c/common/buildcraft/builders/tile/TileQuarry.java#L336

PR #4239, opened 2018-07-27
Cubic Chunks compatibility

EncryptedKitten:
Buildcraft can run with the Cubic Chunks mod without crashing and functions perfectly except some block will not operate below Y=0, such as the Quarry and Mining Well not mining the blocks below Y=0, and the pump not pulling fluids from blocks below Y=0. This adds a config option that sets the bottom of the world  that will default to 0, and functionality will not change but can be set to -1073741824 to allow for Buildcraft mining blocks and the pump block to access all of the Cubic Chunks blocks.


common/buildcraft/core/BCCoreConfig.java
```diff
@@ -196,6 +197,12 @@ public static void preInit(File cfgFolder) {
         propNetworkUpdateRate.setComment(
             "How often, in ticks, should network update packets be sent? Increasing this might help network performance.");
         none.setTo(propNetworkUpdateRate);
+        
+        propWorldMin = config.get(general, "worldMin", 0);
+        propWorldMin.setMinValue(-2147483648).setMaxValue(2147483647);
+        propWorldMin
+            .setComment("How deep is the Minecraft world? (Vanilla = 0, default = 0, Cubic Chuncks = -1073741824)");
```
Generalcamo: This seems a little... odd. Why not use a switch that detects if Cubic Chunks is enabled/loaded and changes it accordingly? ATM there is no reason to set this between 0 and -1073741824.

Generalcamo: This seems a little... odd. Why not use a switch that detects if Cubic Chunks is enabled/loaded and changes it accordingly? ATM there is no reason to set this between 0 and -1073741824.

afdw: It should be possible to use https://github.com/MinecraftForge/MinecraftForge/pull/3919 (which was done specifically for solving this problem in BC).

EncryptedKitten: It has now been updated to automatically set the value based on if the Cubic Chunks mod is loaded. The one instance in which this can cause problems is when Cubic Chunks is a loaded mod but the world is not a Cubic Chunks world (Using an standard world while cubic chunks is installed). One solution to this is not to use standard world while cubic chunks is loaded.

common/buildcraft/lib/misc/CubicChunksChecker.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class CubicChunksChecker {
+
+	/** The Y height level that is the lowest in the world */
+	public static int worldMin = 0;
+
+	public static void postInit()
+	{
+		if(Loader.isModLoaded("cubicchunks")) {
+			worldMin = -1073741824;
```
Barteks2x: That's not how it works

common/buildcraft/factory/tile/TilePump.java
```diff
@@ -90,7 +90,7 @@ private void buildQueue() {
         isInfiniteWaterSource = false;
         Set<BlockPos> checked = new HashSet<>();
         List<BlockPos> nextPosesToCheck = new ArrayList<>();
-        for (BlockPos posToCheck = pos.down(); posToCheck.getY() > 0; posToCheck = posToCheck.down()) {
+        for (BlockPos posToCheck = pos.down(); posToCheck.getY() > CubicChunksChecker.worldMin; posToCheck = posToCheck.down()) {
```
Barteks2x: Assuming the start position is not above the world (which seems reasonable), you could use `world.isOutsideBuildHeight()`.

common/buildcraft/factory/tile/TileMiningWell.java
```diff
@@ -95,7 +95,7 @@ private boolean canBreak() {
     }
 
     private void nextPos() {
-        for (currentPos = pos.down(); currentPos.getY() >= 0; currentPos = currentPos.down()) {
+        for (currentPos = pos.down(); currentPos.getY() >= CubicChunksChecker.worldMin; currentPos = currentPos.down()) {
```
Barteks2x: Assuming the start position is not above the world (which seems reasonable), you could use `world.isOutsideBuildHeight()`.

common/buildcraft/builders/tile/TileQuarry.java
```diff
@@ -358,7 +358,7 @@ public void onPlacedBy(EntityLivingBase placer, ItemStack stack) {
         frameBox.setMin(min);
         frameBox.setMax(max);
         miningBox.reset();
-        miningBox.setMin(new BlockPos(min.getX() + 1, 0, min.getZ() + 1));
+        miningBox.setMin(new BlockPos(min.getX() + 1, CubicChunksChecker.worldMin, min.getZ() + 1));
```
Barteks2x: What about `min.getY() - 256` or some configurable value instead of the constant?

For vanilla, in like 99.9% of the cases, the below 0 value wouldn't matter here because there is bedrock, and in the tiny percentage of cases someone uses it on skylands (which we all know noone does because issue #3610 existed), it will just go some blocks below y=0, which shouldn't really have any bad effects.

And with cubic chunks, it makes no sense to dig deeper than the terrain is generated, because (at least for now) chunks generated by getBlockState won't have any ores.

EncryptedKitten: The problem I noticed when it is -1073741824 on a vanilla world is that after the quarry finishes mining blocks (hits bedrock) some aspects of the game will stop working such as entities freezing, blocks & items cannot be picked up (but blocks can be placed and removed), and the void will not do damage. If the world is reloaded the world will take muck longer to load than usual, locking on 0% for some time until the world loads, and then the effects will go away, but if the quarry is used, such as a new block being placed in the quarry area, and then finishes again, the the same effects will happen.

EncryptedKitten:
It now uses world.isOutsideBuildHeight() to directly determine if a block is usable by a mining well or pump. For the quarry it tests if (0, -1, 0) is outside of the build height it will use 0 as the miningBox min and if it is not it will use -1073741824 for the miningBox min.

Barteks2x:
The problem is that theoretically, cubic chunks can have arbitrary min height. It seems like a much better approach to just go with 256 blocks below where the quarry is.

EncryptedKitten:
I understand that a cubic chunks world could have an arbitrary minimum height, but based on the way Minecraft Forge Vec3i (The class that BlockPos extends) works with signed ints as the variable type, and the Box type for example uses BlockPos for the min/max values, the arbitrary minimum would still be limited to a signed 32 bit integer, that being 2147483647 at a max and -2147483648 as the min.

Generalcamo:
The problem with that approach is that the Quarry will continue to dig down to the minimum value. It would never stop.

afdw:
And if you want to use the min value of int, you probably want `Integer.MIN_VALUE` instead of a literal.

EncryptedKitten:
The only problem with that is it may have to execute unneeded operations for checking the negative blocks of a vanilla world, but that shouldn't have a large effect on gameplay/lag.

EncryptedKitten:
So what should be changed from the current approach with the quarry (checking if negative blocks can exist and if they can setting the miningBox min to -1073741824). The benefits I see in this approach is unchanged functionality in standard worlds and functionality that can take full advantage of current cubic chunks worlds. This approach shouldn't slow down the game, as it only will have to run minimal code at the placement of the quarry.

Barteks2x:
And so far noone seems to see the idea of using min height of 256 blocks below where the quarry is. It would allow to dig full height of vanilla worlds, wouldn't go too insanely deep in vanilla, would work fine in cubicchunks, and wouldn't have the issue of trying to dig infinitely deep.

EncryptedKitten:
What I don't understand is what is the issue of trying to dig to the bottom of the world in both cubic chunks and vanilla, with their respective minimum Y values.

Barteks2x:
And the other reason is that if you don't actually attempt to use cubic chunks specific chunkloading code, you would run into the problem where you are trying to mine ungenerated blocks, and CC would generate them just fine, but population (which spawns ores) is only done on special request (a design choice that makes cascading world generation a non-issue).

Barteks2x:
Actually after looking at this code again, there are more problematic things, I will explain once I'm back home.

AlexIIL:
If it would be better solved by depending on cubic chunks directly (so that we can do certain things properly) then we could always move this into compat.

AlexIIL:
Partially implemented in https://github.com/BuildCraft/BuildCraft/commit/147fac0a7859d0ae4fda9a2923acaf5577b4cdff - but thanks for telling us about this problem!

PR #4271, opened 2018-09-24
More oil config options.

scagood:
#3966

### DIFF
--- a/common/buildcraft/energy/BCEnergyConfig.java
+++ b/common/buildcraft/energy/BCEnergyConfig.java
@@ -39,9 +39,9 @@ public class BCEnergyConfig {
     public static int largeSpoutMinHeight;
     public static int largeSpoutMaxHeight;
 
-    public static Double smallOilGenProb;
-    public static Double mediumOilGenProb;
-    public static Double largeOilGenProb;
+    public static double smallOilGenProb;
+    public static double mediumOilGenProb;
+    public static double largeOilGenProb;
 
     public static final TIntSet excludedDimensions = new TIntHashSet();
     public static final Set<ResourceLocation> excessiveBiomes = new HashSet<>();
@@ -83,7 +83,7 @@ public class BCEnergyConfig {
         );
 
         propOilWellGenerationRate = BCCoreConfig.config.get(
-            "worldgen.oil", "oilWellGenerationRate", 1.0,
+            "worldgen.oil", "generationRate", 1.0,
             "The rate of occurrence of oil wells."
         );
 
@@ -101,8 +101,8 @@ public class BCEnergyConfig {
         );
 
         propEnableOilSpouts = BCCoreConfig.config.get(
-                "worldgen.oil.spouts", "enable", true,
-                "Whether oil spouts are generated or not. The oil spring at the bottom of large lakes will still exist."
+            "worldgen.oil.spouts", "enable", true,
+            "Whether oil spouts are generated or not. The oil spring at the bottom of large lakes will still exist."
         );
 
         propSmallSpoutMinHeight = BCCoreConfig.config.get(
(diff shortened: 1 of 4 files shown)


common/buildcraft/energy/BCEnergyFluids.java
```diff
@@ -110,7 +110,7 @@ private static BCFluid defineFluid(int[] data, int heat, String name) {
             new ResourceLocation(fluidTexture + "_flow"));
         def.setBlockName(name + "_heat_" + heat);
         def.setMapColour(getMapColor(texDark));
-        def.setFlammable(true);
+        def.setFlammable(BCEnergyConfig.enableOilBurn);
```
AlexIIL: I've only just noticed that the current core makes *every* BC fluid burn, which includes residue. Instead can you fix that as well by adding another element to the data array and declare it at the top of the method, like `sticky`? Perhaps only look at the config value once in this method rather than in the data array.

common/buildcraft/energy/BCEnergyConfig.java
```diff
@@ -32,6 +32,17 @@
...
+    public static boolean enableOilBurn;
+
+    public static int smallSpoutMinHeight;
+    public static int smallSpoutMaxHeight;
+    public static int largeSpoutMinHeight;
+    public static int largeSpoutMaxHeight;
+
+    public static Double smallOilGenProb;
+    public static Double mediumOilGenProb;
+    public static Double largeOilGenProb;
```
AlexIIL: Did you mean to make these objects? (Rather than primitives - `double` vs `Double`)

common/buildcraft/energy/BCEnergyConfig.java
```diff
@@ -82,13 +154,13 @@ public static void preInit() {
...
         };
-        propExcludedBiomes = BCCoreConfig.config.get("worldgen", "excludedBiomes", _excluded);
+        propExcludedBiomes = BCCoreConfig.config.get("worldgen.oil", "excludedBiomes", _excluded);
         propExcludedBiomes.setComment("Biome registry names (e.g. 'minecraft:hell','minecraft:jungle')"
             + " of biomes that should never generate oil.");
         world.setTo(propExcludedBiomes);
 
         int[] _dims = { -1, 1 };
-        propExcludedDimensions = BCCoreConfig.config.get("worldgen", "excludedDimensions", _dims);
+        propExcludedDimensions = BCCoreConfig.config.get("worldgen.oil", "excludedDimensions", _dims);
```
AlexIIL: Please don't change the names - you'll break existing configs.

Either add a comment to https://github.com/BuildCraft/BuildCraft/issues/4272 to do this later *or* move the existing config section to the new one. (I haven't done it before but I think it's possible)

common/buildcraft/energy/generation/OilGenerator.java
```diff
@@ -110,13 +114,13 @@ public void generate(Random random, int chunkX, int chunkZ, World world, IChunkG
...
-        if (rand.nextDouble() <= CHANCE_LARGE * bonus) {
+        if (rand.nextDouble() <= BCEnergyConfig.largeOilGenProb * bonus) {
             // 0.04%
             type = GenType.LARGE;
-        } else if (rand.nextDouble() <= CHANCE_MEDIUM * bonus) {
+        } else if (rand.nextDouble() <= BCEnergyConfig.mediumOilGenProb * bonus) {
             // 0.1%
             type = GenType.MEDIUM;
-        } else if (oilBiome && rand.nextDouble() <= CHANCE_LAKE * bonus) {
+        } else if (oilBiome && rand.nextDouble() <= BCEnergyConfig.smallOilGenProb * bonus) {
```
AlexIIL: Maybe get rid of the constants? They won't be used anymore, unless you want to refer to them statically from the config class.

common/buildcraft/energy/BCEnergyConfig.java
scagood: They're now moved I believe

common/buildcraft/energy/BCEnergyFluids.java
AlexIIL: Right sorry. There's a big ugly data array here: 
https://github.com/BuildCraft/BuildCraft/blob/35b4bce36f649421b7ea4a0380db6bcc1f3de543/common/buildcraft/energy/BCEnergyFluids.java#L45 where all of the arguments are taken from

AlexIIL:
Oh, I've just noticed that some of the files you've changed don't have a license :/

Not your fault but I think I'll need to generate those. Oh and ensure that I don't add a license to a file that has been modified by anyone other than myself.

PR #4299, opened 2018-11-09
Add Power Requested trigger

therealfarfetchd:
Trigger turns on when requested power is more than 1 MJ.

### DIFF
--- a/common/buildcraft/transport/statements/TriggerPowerRequested.java
+++ b/common/buildcraft/transport/statements/TriggerPowerRequested.java
@@ -1,22 +1,16 @@
 package buildcraft.transport.statements;
 
-import net.minecraft.util.EnumFacing;
-
-import java.util.Objects;
-import java.util.stream.Stream;
-
 import javax.annotation.Nullable;
 
 import buildcraft.api.core.render.ISprite;
-import buildcraft.api.mj.IMjReceiver;
-import buildcraft.api.mj.MjAPI;
 import buildcraft.api.statements.IStatementContainer;
 import buildcraft.api.statements.IStatementParameter;
 import buildcraft.api.statements.ITriggerInternal;
+import buildcraft.api.transport.pipe.IPipeHolder;
 import buildcraft.core.statements.BCStatement;
 import buildcraft.lib.misc.LocaleUtil;
 import buildcraft.transport.BCTransportSprites;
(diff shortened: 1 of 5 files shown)


AlexIIL:
Erm, can you revert your cleanup comment and just check to see if the requested is greater than 0 for any of them? (As some things request less than 1 MJ I think?)

Although you could probably extract the implementation of `PipeFlowPower.Section.getPowerRequested()` to the main class and provide the `this.side` as null for the trigger, then check and cast the `IPipeHolder.getFlow()` into `PipeFlowPower` rather than the `TilePipeHolder` itself.

(Generally we've found replacing lambdas with normal loops to be quite a decent optimisation)

therealfarfetchd:
The cleanup commit fixes a bug, but I guess I can convert that into a normal loop.

therealfarfetchd:
At first, it cycled through the facings one by one and when it found a capability, it took it. Now if the only power consumer was on that side, getPowerRequested() would return 0 since it filters out the requested power of the side you got the capability from. That is not an issue anymore though

common/buildcraft/transport/statements/TriggerPowerRequested.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class TriggerPowerRequested extends BCStatement implements ITriggerInternal {
+
+    public TriggerPowerRequested() {
+        super("buildcraft:powerRequested");
+    }
+
+    @Override
+    public boolean isTriggerActive(IStatementContainer source, IStatementParameter[] parameters) {
+        final PipeFlowPower flow = (PipeFlowPower) ((IPipeHolder) source.getTile()).getPipe().getFlow();
```
AlexIIL: Sorry about my last bit of advice, clearly I had forgotten how triggers worked :/

```suggestion
        if (!(source instanceof IGate)) {
            return false;
        }
        IPipeFlow f = ((IGate) source).getPipeHolder().getPipe().getFlow();
        if (!(f instanceof PipeFlowPower)) {
            return false;
        }
        final PipeFlowPower flow = (PipeFlowPower) f;
```

AlexIIL:
> build failed with import order violation

Oh well obviously I can't merge this yet then :p

AlexIIL:
Thanks!

PR #4372, opened 2019-03-03
Release ByteBufs that are ensured empty

Chocohead:
Ever since Forge started doing `ByteBuf#retain` on packets, running with `-Dio.netty.leakDetection.level=advanced` to find unreleased packets results in periodic warnings from packets BuildCraft hasn't released. Given that `MessageUtil#ensureEmpty` is for checking that a buffer is empty, it is a reasonable presumption that it isn't going to be used subsequently thus is a good place to release the buffer too.
Releasing buffers can cause issues if they're already released, but from what I've tested this never happens with the way BC is networking right now at the very least.

### DIFF
--- a/common/buildcraft/lib/misc/MessageUtil.java
+++ b/common/buildcraft/lib/misc/MessageUtil.java
@@ -284,7 +284,7 @@ public class MessageUtil {
     }
 
     /** Checks to make sure that this buffer has been *completely* read (so that there are no readable bytes left
-     * over */
+     * over) */
     public static void ensureEmpty(ByteBuf buf, boolean throwError, String extra) {
         int readableBytes = buf.readableBytes();
         int rb = readableBytes;
@@ -339,5 +339,7 @@ public class MessageUtil {
             }
             buf.clear();
         }
+
+        buf.release(); //Finished with the buffer so it can now be deallocated
     }
 }


AlexIIL:
I feel like this would be better in the same place that allocated the byte buffer? Also what version of forge started doing this? The version that BC uses doesn't seem to be logging anything even with leakDetection set to advanced.

Chocohead:
The buffers are allocated within `IMessage`s (or most of them at the very least), so the only way to release them as it stands would be within `finalise` and just expect the messages to be GC'd before the buffers (which is probably a fair presumption). There is at least [one place](https://gist.github.com/Chocohead/4897f975a93bb850c4af9da89a0f3d1d#file-with-pr) where this PR still doesn't cover though, so it's probably worth looking for more catch all solutions.

I'm not sure if Netty's logging doesn't redirect properly normally, but in dev at least loading into a world especially [this](https://gist.github.com/Chocohead/4897f975a93bb850c4af9da89a0f3d1d#file-without-pr) is a fair sample of what gets printed. Potentially using `paranoid` over `advanced` might eke out more. Entering the world especially drives the warnings, as it is from the GC having more pressure on it that will bring the warnings out faster as far as I can tell. BC is quite well behaved to not send packets that frequently which plays a part too, Stubborn Pipe's per tick render update packets that can happen from unstable power flows appear worse than BC's own pipes for example.

I don't remember which Forge version started the warnings, but MinecraftForge/MinecraftForge#4510 added a release in `FMLIndexedMessageToMessageCodec` to the main packet buffer which means not all of BC's packets will trigger warnings. If a part of the packet's buffer is wrapped (such as `MessageUpdateTile#payload`) that won't be released (unlike fully wrapping the buffer) which is where the warnings are spawning from.

AlexIIL:
I *think* that putting them in a finally block is better, however either way this should be fixed now.

PR #4453, opened 2019-08-03
Fix #4407 - Minecraft doesn't crash anymore when gate sets filler pattern

### LINKED ISSUE #4407: Crash with gate trying to set filler pattern
BuildCraft version: 7.99.24.1
Forge version: 14.23.5.2837
Link to crash report or log: https://gist.github.com/128-bit-guy/5e5b9706bfd9925fe9c02674ae44e276
Singleplayer or multiplayer: singleplayer
Steps to reproduce:
1) Place filler
2) Place pipe
3) Add gate to pipe
4) Set gate action to set filler pattern
5) Activate gate

128-bit-guy:
Game doesn't crash anymore, when gate sets filler pattern

### DIFF
--- a/common/buildcraft/builders/tile/TileFiller.java
+++ b/common/buildcraft/builders/tile/TileFiller.java
@@ -496,7 +496,7 @@ public class TileFiller extends TileBC_Neptune
     @Override
     public void setPattern(IFillerPattern pattern, IStatementParameter[] params) {
         patternStatement.set(pattern);
-        IntStream.range(0, patternStatement.maxParams).forEach(i -> patternStatement.set(i, params[i]));
+        IntStream.range(0, params.length).forEach(i -> patternStatement.set(i, params[i]));
         finished = false;
         lockedTicks = 3;
     }


common/buildcraft/builders/tile/TileFiller.java
```diff
@@ -496,7 +496,7 @@ public IBox getBox() {
     @Override
     public void setPattern(IFillerPattern pattern, IStatementParameter[] params) {
         patternStatement.set(pattern);
-        IntStream.range(0, patternStatement.maxParams).forEach(i -> patternStatement.set(i, params[i]));
+        IntStream.range(0, params.length).forEach(i -> patternStatement.set(i, params[i]));
```
AlexIIL: Should this be `Math.min(params.length, patternStatement.maxParams)` instead?

AlexIIL: Actually never mind - I've noticed a few other places where this was wrong so I'll fix those too.

128-bit-guy: I don't think, that it should, because filler action won't have more parameters, than patternStatement.maxParamenters.

AlexIIL: Technically the params array comes from the gate length, not the filler action's maximum number of parameters. (However buildcraft itself doesn't have any gates with more than 3 paramaters so this shouldn't actually ever happen)

PR #4555, opened 2020-04-11
Fix #4553

### LINKED ISSUE #4553: Diamond Pipe doesn't distinguish cables from ic2
- Minecraft Version: 1.12.2
 - Forge Version: 14.23.5.2847
 - BuildCraft Version: buildcraft-all-7.99.24.5
 - IndustrialCraft Version: industrialcraft-2-2.8.176-ex112

1. Create a diamond pipe.
2. Attach two cobblestone pipes for output.
3. Attach some pipe for input.
4. In diamond pipe filters, set "Copper Cable" for the first pipe, and "Insulated Copper Cable" for the second.
5. Pass multiple cables into the diamond pipe.

Expected: The "Copper Cable" must go out from the first pipe, and the "Insulated Copper Cable" - from the second pipe. 
Actual: The items choose pipes randomly. 

This also works with "Tin Cable".

ashlanderr:
It is a part of the fix. The second part will go to BuildCraftCompat. There will be a registration of NBT keys for ic2 cables.

### DIFF
--- a/common/buildcraft/lib/misc/StackMatchingPredicate.java
+++ b/common/buildcraft/lib/misc/StackMatchingPredicate.java
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) 2020 SpaceToad and the BuildCraft team
+ * This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not
+ * distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/
+ */
+
+package buildcraft.lib.misc;
+
+import javax.annotation.Nonnull;
+
+import net.minecraft.item.ItemStack;
+
+@FunctionalInterface
+public interface StackMatchingPredicate {
+    boolean isMatching(@Nonnull ItemStack base, @Nonnull ItemStack comparison);
+}
--- a/common/buildcraft/lib/misc/StackUtil.java
+++ b/common/buildcraft/lib/misc/StackUtil.java
@@ -6,8 +6,12 @@
 
 package buildcraft.lib.misc;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
@@ -17,6 +21,7 @@ import javax.annotation.Nullable;
 
 import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.util.NonNullList;
@@ -35,6 +40,11 @@ public class StackUtil {
     // Actually the entire MC
     @Nonnull
     public static final ItemStack EMPTY;
+
+    /**
+     * Registry of additional rules for {@link #isMatchingItem}.
+     */
+    private static final Map<Item, List<StackMatchingPredicate>> matchingPredicates = new HashMap<>();
 
     static {
         ItemStack stack = ItemStack.EMPTY;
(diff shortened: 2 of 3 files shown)


common/buildcraft/lib/misc/StackUtil.java
```diff
@@ -298,6 +304,12 @@ public static boolean isMatchingItem(@Nonnull final ItemStack base, @Nonnull fin
         if (base.getItem() != comparison.getItem()) {
             return false;
         }
+
+        Set<String> matchingNBT = matchingNBTRegistry.get(base.getItem());
+        if (matchingNBT != null && !isNbtSubsetEquals(base, comparison, matchingNBT)) {
+            return false;
+        }
```
AlexIIL, line 311: Can you move this to an `else` statement below `if (matchNBT)` so that we still perform strict comparison first?

common/buildcraft/lib/misc/StackUtil.java
```diff
@@ -6,9 +6,7 @@
 
 package buildcraft.lib.misc;
 
-import java.util.Collection;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
AlexIIL, line 9: Don't use star imports.

common/buildcraft/lib/misc/StackUtil.java
```diff
@@ -17,10 +15,11 @@
 
 import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTBase;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.util.NonNullList;
-
```
AlexIIL, line 23: Please don't remove import spacing.

(There's a list of import order here: https://github.com/BuildCraft/BuildCraft/blob/8.0.x-1.12.2/guidelines/buildcraft.checkstyle#L15, where different sections should be separated).

common/buildcraft/lib/misc/StackUtil.java
```diff
@@ -298,6 +304,12 @@ public static boolean isMatchingItem(@Nonnull final ItemStack base, @Nonnull fin
         if (base.getItem() != comparison.getItem()) {
             return false;
         }
+
+        Set<String> matchingNBT = matchingNBTRegistry.get(base.getItem());
+        if (matchingNBT != null && !isNbtSubsetEquals(base, comparison, matchingNBT)) {
+            return false;
+        }
+
```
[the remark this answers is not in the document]
ashlanderr: If I will change the NBT set to a predicate, then this condition and the `matchNBT` are not exactly the opposite. The predicate can match for anything in `ItemStack`, for example, a name of an item. I think it would be better to place my code not in the else branch but after the `matchingNBT` block.

common/buildcraft/lib/misc/StackUtil.java
ashlanderr: Sorry, I didn't notice that. IDEA automatically changed it.

common/buildcraft/lib/misc/StackUtil.java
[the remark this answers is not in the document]
ashlanderr: As I understand, if NBTs are equals, then there is no other way to tell the difference between item stacks. So, yes, it must be in else block to prevent redundant checks. But maybe I'm wrong, I am not good at Minecraft API.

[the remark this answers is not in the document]
AlexIIL: Yes, ItemStack's only have three fields that we care about: Item, metadata/damage (int), and NBT.

(Items also can have capabilities but as BC doesn't handle them at all it's probably not a good idea to start caring about them in this PR).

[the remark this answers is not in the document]
AlexIIL: Thanks!

common/buildcraft/lib/misc/StackMatchingPredicate.java
```diff
@@ -0,0 +1,10 @@
+package buildcraft.lib.misc;
```
AlexIIL, line 1: One last thing: can you add the buildcraft license header to the top of these two files? (Which is currently at the top of StackUtil)

ashlanderr:
I already have created a [branch](https://github.com/ashlanderr/BuildCraftCompat/tree/8.0.x-1.12.2-ic2-cables-matching) in BuildCraftCompat with the second part of the fix. I will send a new PR when this PR will be merged.

AlexIIL:
Thanks!

PR #4618, opened 2020-11-21
Allow oredicted ItemStacks with stacksizes greater than 1

glowredman:
Affects `buildcraft.core.recipes.FlexibleRecipe#setContents()`.

This change is backwardscompatible and fullfills this promise:
grafik

I was not able to build the jars, even with a fresh clone. It always crashes at `:javadoc`. Maybe this [log](https://gist.github.com/glowredman/59c3fbd6c1fe1cb963c22b5a69c92177) of `gradlew build --stacktrace --info --debug` can help with that...
When I removed the javadoc tasks, after that, it built as expected.

### DIFF
--- a/common/buildcraft/core/recipes/FlexibleRecipe.java
+++ b/common/buildcraft/core/recipes/FlexibleRecipe.java
@@ -124,7 +124,13 @@ public class FlexibleRecipe<T> implements IFlexibleRecipe<T>, IFlexibleRecipeVie
 		energyCost = iEnergyCost;
 		craftingTime = iCraftingTime;
 
+		boolean skip = false;
 		for (int index = 0; index < input.length; index++) {
+			if (skip) {
+				skip = false;
+				continue;
+			}
+			
 			Object i = null;
 			Object ii = input[index];
 			if (ii == null) {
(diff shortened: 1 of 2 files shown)


asiekierka:
You can do this a little cleaner... I'll do it, one second.

common/buildcraft/core/recipes/FlexibleRecipe.java
```diff
@@ -145,7 +152,19 @@ public void setContents(String iid, Object iioutput, int iEnergyCost, long iCraf
 			} else if (i instanceof List) {
 				inputItemsWithAlternatives.add((List) i);
 			} else if (i instanceof String) {
-				inputItemsWithAlternatives.add(OreDictionary.getOres((String) i));
+				if (index + 1 >= input.length) {
+					inputItemsWithAlternatives.add(OreDictionary.getOres((String) i));
+				} else if (input[index + 1] instanceof Integer) {
+					skip = true;
```
asiekierka, line 158: Could you please simply use `index++;` here instead of `skip = true;`? That's much less hacky, and you can do this with a for loop.

asiekierka (changes requested):
Other than this, looks good to me - and will be merged. Though I can give no promises on an official update... long out of my jurisdiction.

common/buildcraft/core/recipes/FlexibleRecipe.java
glowredman: I did, but the style checker complained aborted the build, I can see if there is an annotation for it to ignore that method though.

asiekierka: It's 1.7.10, nobody supports it anyway, I wouldn't normally advise this but just throw out the style checker and Javadoc if it's causing problems... unless @AlexIIL wants to backport some improvements from the new branches.

asiekierka:
@AlexIIL feel free to merge whenever, and/or do something with the checkstyle, this is good to me

PR #4623, opened 2020-12-10
Fix #4608 Mining well can not move through flowing water blocks

### LINKED ISSUE #4608: Mining Well cannot mine through flowing water blocks.
**Please complete the following information:**
 - Forge Version: 1.12.2 - 14.23.5.2854
 - BuildCraft Version : buildcraft-all-7.99.24.6
 - Any additonal mods: 
    - extrautils2-1.12-1.9.9
    - projectred-all-1.12.2-4.9.120

**To Reproduce**
1. Create New World
2. Place power production and mining well setup
3. Place flowing water several blocks under mining well
4. Observe

Expected behavior:  
> The mining well works it's way through the flowing water mining the blocks underneath in the same fashion as the quarry does.

Actual behavior:  
> The mining well stops advancing when the mining pipe reaches the flowing water block. It continues drawing small amounts of power until the flowing water block is removed, in which case it continues normal operation.

Smartheads:
### Changed block breaking criteria to match Quarry

Changed criteria on which it's decided, if the mining well can break a block to match the quarry. This should resolve issue #4608, in which the mining well can not move through flowing water blocks, since the quarry doesn't suffer from such issue.

**Please test proposed code before accepting the pull request.**

### DIFF
--- a/common/buildcraft/factory/tile/TileMiningWell.java
+++ b/common/buildcraft/factory/tile/TileMiningWell.java
@@ -87,7 +87,7 @@ public class TileMiningWell extends TileMiner {
     }
 
     private boolean canBreak() {
-        if (world.isAirBlock(currentPos) || BlockUtil.isUnbreakableBlock(world, currentPos, getOwner())) {
+        if (world.getBlockState(currentPos).getBlockHardness(world, currentPos) < 0) {
             return false;
         }
 


Aitormantilla:
I want to fix this bug in my world but i dont known how to change the code. Can you send the mod fixed? I have been trying to fix it but I didnt found anything. Thanks

PR #4656, opened 2021-10-28
Fix NullPointerException crash in PipeFlowFluids

anonyco:
FIXES a bug where accesing lastSentDirection.nbtValue throws a NullPointerException when placing a wooden fluid pipe next to a Magneticraft Infinite Water block and then breaking wooden fluid pipe. To the best of my Java ability I couldn't figure out why :(

### DIFF
--- a/common/buildcraft/transport/pipe/flow/PipeFlowFluids.java
+++ b/common/buildcraft/transport/pipe/flow/PipeFlowFluids.java
@@ -776,13 +776,16 @@ public class PipeFlowFluids extends PipeFlow implements IFlowFluid, IDebuggable 
     class Section implements IFluidHandler {
         final EnumPipePart part;
 
-        int amount;
+        int amount = 0;
 
-        int lastSentAmount;
+        int lastSentAmount = 0;
 
-        Dir lastSentDirection;
+        /* Accesing lastSentDirection.nbtValue throws a NullPointerException when placing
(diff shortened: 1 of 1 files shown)


anonyco:
I have fixed my PR so that it now removes those two lines as you suggested.

AlexIIL:
You're nearly correct - with one exception: you can only get a NullPointerException when dereferencing a value with null - which only happens when you use `.` on a variable/field/*method result* containing null. Using the identity comparison operators (`==` or `!=`) on null is perfectly safe. As such `lastSentDirection` being null is actually always safe, since the only place it was dereferenced is when it was saved to disk - which needs to be removed. It's also completely harmless (*and probably beneficial*) to initialise it though - since `Dir.NONE` is what the client starts out with.

AlexIIL:
Basically this looks good now, thanks!

anonyco:
> You're nearly correct - with one exception: you can only get a NullPointerException when dereferencing a value with null - which only happens when you use `.` on a variable/field/_method result_ containing null. Using the identity comparison operators (`==` or `!=`) on null is perfectly safe. As such `lastSentDirection` being null is actually always safe, since the only place it was dereferenced is when it was saved to disk - which needs to be removed. It's also completely harmless (_and probably beneficial_) to initialise it though - since `Dir.NONE` is what the client starts out with.

Thank you for taking the time to explain this to me. I learn something new about Java every day lol.

PR #4688, opened 2022-10-25
[1.7.10] Fix Compat with NEI Unofficial

DrParadox7:
Fixes `java.lang.NumberFormatException' caused by chat formatting when using NEI Unofficial 2.19+ from https://github.com/GTNewHorizons/NotEnoughItems/issues/310.

@asiekierka Would you be able to merge this?

### DIFF
--- a/common/buildcraft/builders/gui/GuiBlueprintLibrary.java
+++ b/common/buildcraft/builders/gui/GuiBlueprintLibrary.java
@@ -8,6 +8,7 @@
  */
 package buildcraft.builders.gui;
 
+import net.minecraft.util.EnumChatFormatting;
 import org.lwjgl.opengl.GL11;
 
 import net.minecraft.client.gui.GuiButton;
@@ -62,7 +63,7 @@ public class GuiBlueprintLibrary extends GuiBuildCraft {
 				break;
 			}
 			LibraryId bpt = library.entries.get(i);
-			String name = bpt.name;
+			String name = EnumChatFormatting.getTextWithoutFormattingCodes(bpt.name);
 
 			if (name.length() > DefaultProps.MAX_NAME_SIZE) {
 				name = name.substring(0, DefaultProps.MAX_NAME_SIZE);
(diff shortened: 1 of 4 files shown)


DrParadox7:
I've updated the buildscript. 
This should fix the Github Action checks.

common/buildcraft/builders/gui/GuiBlueprintLibrary.java
```diff
@@ -62,7 +63,7 @@ protected void drawGuiContainerForegroundLayer(int par1, int par2) {
 				break;
 			}
 			LibraryId bpt = library.entries.get(i);
-			String name = bpt.name;
+			String name = EnumChatFormatting.getTextWithoutFormattingCodes(bpt.name);
```
asiekierka, line 66: Why here? How would this field be affected by NEI?

common/buildcraft/core/list/ListOreDictionaryCache.java
```diff
@@ -42,6 +43,7 @@ private void addToNamingCache(String s, int id) {
 	public static String getType(String name) {
 		// Rules for finding type:
 		// - Split just before the last uppercase character found.
+		name = EnumChatFormatting.getTextWithoutFormattingCodes(name);
```
asiekierka, line 46: Why here? This should operate on OreDictionary strings only.

common/buildcraft/silicon/render/PackageFontRenderer.java
```diff
@@ -33,6 +34,7 @@ public int getStringWidth(String s) {
 
 	@Override
 	public int drawString(String s, int x, int y, int color, boolean shadow) {
+		s = EnumChatFormatting.getTextWithoutFormattingCodes(s);
```
asiekierka, line 37: NEI shouldn't be affecting these lines, but that's a tolerable patch; either way, this should run *after* the `realRenderer.drawString` check/call, because otherwise it will strip formatting codes from lines which are not affected by the code below.

asiekierka:
Please try this one.

asiekierka:
I guess, if GT:NH is merging it...

PR #4756, opened 2026-07-02
Hey BC team, We want to introduce to you BuildCraft Community Edition

shipovskijkorp:
Hi BC devs, I would like you to take a look at my port and rate it. This port is based on BC8 and BC7. There are even robots (16/17). I rate the port as 99% of the original content and have already published it. And I would like you to rate my work. 
With love to BuildCraft and respect to you from CurativeTree, shipovskijkorp and other guys from community.


shipovskijkorp:
I think to do many test, and after port it to 1.20.1.

PR #4764, opened 2026-08-15
[1.12.2] Fix lasers searching for targets when unnecessary

oddkibble:
While profiling a server, I noticed lasers (N < 50) were responsible for a lot of the tick time, spending it searching for targets. Because some surrounding blocks were updating every tick, `setWorldUpdated` got triggered just as often, making the lasers search for targets every time.
As far as I could tell though, all that matters is whether the block material changed: specifically, whether a block started/stopped blocking the laser, or started/stopped being a viable target.
Adding a check for it made lasers disappear from subsequent profiles (and didn't seem to break anything).

### DIFF
--- a/common/buildcraft/silicon/tile/TileLaser.java
+++ b/common/buildcraft/silicon/tile/TileLaser.java
@@ -85,7 +85,9 @@ public class TileLaser extends TileBC_Neptune implements ITickable, IDebuggable,
 
     @Override
     public void setWorldUpdated(World world, BlockPos eventPos, IBlockState oldState, IBlockState newState, int flags) {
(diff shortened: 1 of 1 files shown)


Aedial:
I would recommend to still scan for ILaserTargetBlock updates, to avoid some cases like #4768 getting ignored by mistake. There should be few of these blocks in each update cone.

Aedial:
Also, it has the unfortunate consequence of not re-scanning when the laser is rotated. You need to add an additional check if :
- The block being updated is the laser tile itself
- The old state has a different rotation property than the new state

In which case, you still update

PR #4768, opened 2026-09-14
Fix lasers not finding table if it loads after them

Aedial:
If the table loads after the lasers, they might not realize a target appeared and never "wake up", causing them to just stop working. This PR sends a "changed" signal to the lasers' listener after it initialized.


FlashyGuitar603:
I will kiss you right now, I've had to deal with this BS for so long.

